Issue #104, opened 2025-06-16
Add instance name to admin vm database


eberrigan:
![Image](https://github.com/user-attachments/assets/902eee4f-da27-45f9-b506-8a85ee1cfd03)

It would be useful if this was more similar to the amazon ec2 instance table

PR #125, opened 2025-07-02
Add Rows to the Instance Table for Admin

### LINKED ISSUE #107: Add Row Numbers to the Instances Page
It would be better to count the rows, since admin has to determine how many VMs are available to the clients.

7174Andy:
This PR fixes #107 and fixes https://github.com/talmolab/lablink/issues/104

Screenshot 2025-07-02 at 11 08 27

### DIFF
--- a/lablink-allocator/lablink-allocator-service/conf/config.yaml
+++ b/lablink-allocator/lablink-allocator-service/conf/config.yaml
@@ -9,7 +9,7 @@ db:
 
 machine:
   machine_type: "g4dn.xlarge"
-  image: "ghcr.io/talmolab/lablink-client-base-image:linux-amd64"
+  image: "ghcr.io/talmolab/lablink-client-base-image:linux-amd64-test"
   ami_id: "ami-067cc81f948e50e06"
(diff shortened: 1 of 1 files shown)


lablink-client-base/lablink-client-service/lablink_client_service/conf/config.yaml
```diff
@@ -1,3 +1,3 @@
 allocator:
   host: "localhost"
-  port: 5000
\ No newline at end of file
+  port: 80
```
eberrigan, line 3: why wasn't this an issue before? is this necessary?

eberrigan (approved):

lablink-client-base/lablink-client-service/lablink_client_service/conf/config.yaml
7174Andy: It was not a problem because we used the image in this repo: https://github.com/talmolab/sleap-cuda-container. The image had a `startup.sh` that sets the port to 80.

PR #153, opened 2025-08-12
Prevent Continuous POST API Calls to the Allocator Server

### LINKED ISSUE #147: Prevent Continuous POST Request to the Allocator When Checking GPU
Currently, the `lablink-client-service` sends a POST request every interval although the health status is not changing. Like `udpate_inuse_status.py`, we should send API request when there is a status change, instead of regularly.

7174Andy:
This PR fixes #147. In the `check_gpu.py`, the POST API call is made whenever there are any changes in the status.


eberrigan (approved):
tests please!

PR #172, opened 2025-10-01
Make OIDC in GitHub Actions Configurable Using GitHub Secrets

7174Andy:


eberrigan (changes requested):
Add to the README that this secret needs to be added to the github repo before use

PR #205, opened 2025-10-10
Make Startup script for Client Images Configurable and Fix Minor Bugs

### LINKED ISSUE #191: Pass in `startup.sh` in Run Time for Client Image
Currently, the lablink-client-base-image copies the startup.sh from local repo during build time. However, this restricts the instructors to customize which package to use when creating this as a template. Currently, we only have github repo configurable for cloning. Like what we did for config.yaml to pass during run time, we could take a similar approach to transfer the startup.sh file in the template and run this bash script during run time of the Docker image (mounting to a volume). 

Here is the original code: 
https://github.com/talmolab/lablink/blob/1df4ff7f16a347d019ab6212c6a7abe7caa110ee/lablink-client-base/lablink-client-base-image/Dockerfile#L147-L149

7174Andy:
## Summary

This PR fixes #191. This PR also fixes the unnecessary log of `psql:/app/init.sql:4: NOTICE:  role "lablink" does not exist, skipping` that might confuse the end users of LabLink. This PR also fixes the frontend inconsistencies with the config. 

##  Key Changes

   - **Dynamic Startup Script:**
     - The client's start.sh script is no longer baked into the Docker image.
     - It's now passed from the lablink-infrastructure module to the client container at runtime.
     - This allows for environment-specific startup configurations without rebuilding the client image.

   - **Terraform and `user_data` Updates:**
     - main.tf now includes a startup_script_path variable to specify the script's location.
     - The user_data.sh script in the allocator now creates the startup script on the EC2 instance and mounts it
       into the client container.

   - **Dockerfile Modifications:**
     - The client Dockerfile and Dockerfile.dev have been updated to use a START_SCRIPT environment variable,
       which points to the dynamically mounted startup script.
     - The ENTRYPOINT is now set to /bin/bash -lc, and the CMD executes the START_SCRIPT.

   - **Configuration and Script Relocation:**
     - The packages/client/start.sh has been moved to lablink-infrastructure/config/startup.sh and now serves as
       the default startup script.
     - A custom message has been added to the startup script to confirm its execution.

   - **Bug Fixes**
     - Modified `generate_init_sql.py` so that it suppresses the logs in order to suppress unnecessary NOTICE logs
     - Modified the admin page so that the frontend page aligns with the config files
    

##  Benefits:

   - **Flexibility**: Easily customize client container startup for different environments (dev, test, prod) without
     image modifications.
   - **Centralized Configuration**: Infrastructure-related startup logic is now managed within the
     lablink-infrastructure module.
   - **Simplified Client Image**: The client Docker image is now more generic and less tied to specific startup
     procedures.

### Note: 
- Tested with Cellpose using uv. The software installed successfully in the client VM. 
- Added all the changes in the infrastructure in this PR for lablink-template repo (https://github.com/talmolab/lablink-template/pull/10)

### DIFF
--- a/lablink-infrastructure/main.tf
+++ b/lablink-infrastructure/main.tf
@@ -159,7 +159,7 @@ resource "aws_instance" "lablink_allocator_server" {
     CONFIG_CONTENT        = file("${path.module}/${var.config_path}")
     DOMAIN_NAME           = local.fqdn
     SSL_STAGING           = local.ssl_staging
-    CLIENT_STARTUP_SCRIPT = file("${path.module}/config/start.sh")
+    CLIENT_STARTUP_SCRIPT = file("${path.module}/config/startup.sh")
   })
 
   tags = {


eberrigan (changes requested):
Hi Andrew, thanks for putting time into this, but I assumed this was doing something else. So sorry but we really can't move necessary code for the Docker containers anywhere else and make it optional. If we want a configurable startup script it should be an additional script added via the infrastructure repo config (can make the path to the script configurable using a new config key), but all the containers need to be usable without this script. Think about when the custom script should run without breaking anything else. Let's also add some tests to this repo that could be run in the lablink-images ci verification step.

```
startup_script:
  enabled: true          # default: false if omitted
  path: config/custom-startup.sh   # default: config/custom-startup.sh
  on_error: continue     # options: continue | fail   (default: continue)
```
or hardcode the path to always be the same.

Here is a summary of how it should work via chatGPT

<html>
<body>

<hr>
<h3>💬 PR #205 Review Summary</h3>
<p>You’re thinking about flexibility, which is great. However, the current implementation changes how the startup process fundamentally works and introduces unnecessary complexity. Let’s simplify and align it with our design principles.</p>
<hr>
<h3>🚨 Critical Fix</h3>
<p><strong>Dockerfile CMD syntax:</strong></p>
<pre><code class="language-Dockerfile">CMD ["exec ${START_SCRIPT}"]  # ❌ This won't expand variables
CMD exec ${START_SCRIPT}      # ✅ Use shell form for variable expansion
</code></pre>
<hr>
<h3>⚙️ Core Design Issue</h3>
<p>Right now, the PR makes the <strong>entire startup script configurable</strong> and moves it to the infrastructure repo.<br>
This creates redundancy and breaks our convention-based design.</p>
<h4>Why that’s a problem:</h4>
<ul>
<li>
<p>Duplicates information already defined in <code inline="">config.yaml</code></p>
</li>
<li>
<p>Breaks local development (no default startup script baked into the image)</p>
</li>
<li>
<p>Makes core functionality optional when it shouldn’t be</p>
</li>
</ul>
<hr>
<h3>✅ What We Actually Need</h3>
<p>We only need to <strong>support an optional custom startup script.</strong><br>
The base startup logic should remain unchanged.</p>
<h4>Minimal Implementation:</h4>

### 1) Extend `config/config.yaml`

```yaml
startup_script:
  enabled: true                 # default false if omitted
  path: config/custom-startup.sh  # default path if omitted
  on_error: continue            # continue | fail  (default continue)
```

### 2) `main.tf` — load config and derive locals

> Use `path.root` for user-provided files (deployment repo), and `path.module` for files that live inside the module. Since `custom-startup.sh` is user-owned, favor `path.root`.

```hcl
locals {
  # Centralized config
  config_file = yamldecode(file("${path.root}/config/config.yaml"))

  # Startup script settings (all from config)
  startup_enabled   = try(local.config_file.startup_script.enabled, false)
  startup_path      = try(local.config_file.startup_script.path, "config/custom-startup.sh")
  startup_on_error  = try(local.config_file.startup_script.on_error, "continue")

  # Only read the script content when enabled AND file exists
  startup_script_content = (
    local.startup_enabled && fileexists("${path.root}/${local.startup_path}")
  ) ? file("${path.root}/${local.startup_path}") : ""
}

resource "aws_instance" "lablink_vm" {
  # ...existing config...

  user_data = templatefile("${path.module}/user_data.sh", {
    custom_startup_script = local.startup_script_content
    startup_enabled       = local.startup_enabled
    startup_on_error      = local.startup_on_error
    # ...other existing vars...
  })
}
```

> If you truly want the YAML *inside the module* (as you sketched with `${path.module}/config/config.yaml`), you can keep that—just switch both occurrences of `path.root` above to `path.module`. Most teams prefer `path.root` so deployers can own `config/` without forking the module.

### 3) `user_data.sh` — write script only when enabled + present

```bash
mkdir -p /etc/lablink-client-config

if [ "${startup_enabled}" = "true" ] && [ -n "${custom_startup_script}" ]; then
  echo ">> Custom startup: enabled; writing script"
  cat <<'EOF' > /etc/lablink-client-config/custom-startup.sh
${custom_startup_script}
EOF
  chmod +x /etc/lablink-client-config/custom-startup.sh
else
  echo ">> Custom startup: disabled or not provided"
fi

docker run -dit $DOCKER_GPU_ARGS \
  --env STARTUP_ON_ERROR="${startup_on_error}" \
  --mount type=bind,src=/etc/lablink-client-config,dst=/docker_scripts/,ro \
  # ...rest...
```

### 4) `packages/client/start.sh` — respect policy

```bash
if [ -f /docker_scripts/custom-startup.sh ]; then
  echo ">> Running custom startup script..."
  bash /docker_scripts/custom-startup.sh
  rc=$?
  if [ $rc -ne 0 ]; then
    echo "Warning: custom startup exit code $rc"
    if [ "${STARTUP_ON_ERROR}" = "fail" ]; then
      echo ">> Policy is 'fail' — exiting"
      exit $rc
    fi
  fi
else
  echo ">> No custom startup script present (or disabled)"
fi
```

<p><strong>Template Repo</strong><br>
Add an example script:<br>
<code inline="">config/custom-startup.sh.example</code><br>
This gives users a pattern to follow without modifying core behavior.</p>
</li>
</ol>
<hr>
<h3>📂 Standard Path Convention</h3>

Location | Purpose
-- | --
config/custom-startup.sh | Optional user script (Terraform checks)
/etc/lablink-client-config/custom-startup.sh | Created on EC2 host
/docker_scripts/custom-startup.sh | Mounted into container (read-only)


<p>Everything else stays the same.</p>
<hr>
<h3>🧩 Keep These Good Additions</h3>
<p>✅ PostgreSQL log suppression<br>
✅ Template extension configurability<br>
✅ Removing unused <code inline="">flask db upgrade</code><br>
✅ <code inline="">DEBIAN_FRONTEND</code> and <code inline="">NEEDRESTART_MODE</code> exports</p>
<hr>
<h3>🚫 Remove or Revert These</h3>
<p>❌ Making the startup script fully configurable<br>
❌ Adding variables for startup path (use locals)<br>
❌ Adding startup scripts to <code inline="">lablink-infrastructure/config</code><br>
❌ Removing default <code inline="">start.sh</code> from Docker images</p>
<hr>
<h3>💡 Key Principles</h3>
<ul>
<li>
<p><strong>Convention over configuration</strong> — hardcode standard paths</p>
</li>
<li>
<p><strong>Fail gracefully</strong> — missing custom script should not break startup</p>
</li>
<li>
<p><strong>Optional by design</strong> — core functionality always runs</p>
</li>
<li>
<p><strong>No duplication</strong> — don’t repeat info from <code inline="">config.yaml</code></p>
</li>
</ul>
<hr>
<h3>🧾 Summary</h3>
<p>The current PR solves a real problem but goes too far.<br>
We just need:</p>
<ul>
<li>
<p>Optional custom script support via <code inline="">fileexists()</code></p>
</li>
<li>
<p>Keep default startup intact</p>
</li>
<li>
<p>Provide an example in the template repo</p>
</li>
<li>
<p>Fix the Docker CMD bug</p>
</li>
</ul>
<p>Once these changes are made, the implementation will be clean, minimal, and fully aligned with LabLink’s conventions.</p>
<hr>

PR #211, opened 2025-10-15
Add a Custom Startup Script Feature without Affecting the Core Startup script

7174Andy:
This PR will close #205. This PR adds a new feature of configurable startup script as discussed in PR #205. 

## Overview of how this works

1. `custom-startup.sh` file is defined in the [template repo](https://github.com/talmolab/lablink-template)

2. The infrastructure Terraform transfers the custom startup script to the allocator Docker container ([Reference](https://github.com/talmolab/lablink-template/blob/main/lablink-infrastructure/main.tf))

3. Then, the allocator's Terraform script for client VMs are ran to transfer the custom startup script: 

```hcl
startup_content = fileexists(var.custom_startup_script_path) ? file(var.custom_startup_script_path) : ""
```

4. The `user_data.sh` will transfer the startup script to the client's Docker container for use

```bash
cat <<'EOF' > /etc/config/custom-startup.sh
${startup_content}
EOF

chmod +x /etc/config/custom-startup.sh

echo ">> Starting container..."
if docker run -dit $DOCKER_GPU_ARGS \
    --mount type=bind,src=/etc/config,dst=/docker_scripts/,ro \
    -e ALLOCATOR_HOST="${allocator_ip}" \
    -e ALLOCATOR_URL="${allocator_url}" \
    -e TUTORIAL_REPO_TO_CLONE="${repository}" \
    -e VM_NAME="lablink-vm-${resource_suffix}-${count_index}" \
    -e SUBJECT_SOFTWARE="${subject_software}" \
    -e CLOUD_INIT_LOG_GROUP="${cloud_init_output_log_group}" \
    -e AWS_REGION="${region}" \
    -e STARTUP_ON_ERROR="${startup_on_error}" \
    --network host \
    "${image_name}"; then
    send_status "running"
```

5. The `start.sh` within the Docker container will run the custom startup script with all the fundamental services starting.

### DIFF
--- a/packages/allocator/src/lablink_allocator_service/conf/structured_config.py
+++ b/packages/allocator/src/lablink_allocator_service/conf/structured_config.py
@@ -161,9 +161,9 @@ class AllocatorConfig:
 class StartupConfig:
     """Configuration for startup behavior of the allocator service.
     Attributes:
-        startup_enabled (bool): Whether startup script execution is enabled.
-        startup_path (str): Path to the startup script to be executed.
-        startup_on_error (str): Behavior on startup script error. Options:
+        enabled (bool): Whether startup script execution is enabled.
+        path (str): Path to the startup script to be executed.
+        on_error (str): Behavior on startup script error. Options:
             - "continue": Log the error and continue startup.
             - "fail": Abort startup on error.
     """


packages/allocator/src/lablink_allocator_service/terraform/user_data.sh
```diff
@@ -131,15 +131,27 @@ if [ "$HAS_GPU" = true ]; then
...
 fi
 
+echo "> Creating config directory…"
+sudo mkdir -p /etc/config
+
+cat <<'EOF' > /etc/config/startup.sh
+${startup_content}
+EOF
+
+chmod +x /etc/config/startup.sh
```
eberrigan: I am a bit confused with the similar names for the files, double check that these file names are correct, let's try to have consistent file naming schemes (start.sh for Docker entrypoint, user_data.sh for terraform starter scripts for VMs, and custom_start.sh for the custom script)

eberrigan (changes requested):
We can add some tests with typical startup scripts as fixtures like installing a package or something like that in either ci or lablink-images ci. 

Is the diff coverage by python tests? 

Make sure the paths are correct and consistent throughout for ease in understanding

.github/workflows/lablink-images.yml
```diff
@@ -385,3 +385,35 @@ jobs:
...
+
+      - name: Verify custom-startup.sh execution
+        run: |
+          echo "Testing custom-startup.sh execution..."
+          CUSTOM_STARTUP_SCRIPT_CONTENT=$(cat <<'EOF'
+          #!/bin/bash
+          echo "CUSTOM_STARTUP_SCRIPT_EXECUTED_SUCCESSFULLY"
+          sudo apt-get update -y
+          sudo apt-get install -y jq
+          echo "JQ_VERSION: $(jq --version)"
```
eberrigan: could keep a few different scripts in the tests folder maybe in a test_data folder and have a separate CI to try different typical test scripts. We could keep this CI shorter but we don't need to necessarily since it has stages and we already have the shorter CI testing.

eberrigan (commented):
some suggestions for improved testing but my main concern is `startup_content = fileexists("/config/custom-startup.sh") ? file("/config/custom-startup.sh") : ""` because i thought that local should be defined from the config. Lmk if I misunderstood that.

PR #236, opened 2025-11-10
Change uv Installation Methods and Add More Packages to the Client

7174Andy:
## Summary

This PR modernizes the UV package installation approach in both the allocator and client Docker images, and adds additional utility packages to the client image for improved user experience.

### 1. UV Installation Method Updates

**Allocator (`packages/allocator/Dockerfile`)**:
- Replaced direct `uv pip install` approach with UV's project-based workflow
- Now uses `uv init --python 3.11 --no-readme` to initialize a UV project
- Creates virtual environment with `uv venv /app/.venv`
- Uses `uv add lablink-allocator-service==${PACKAGE_VERSION} --no-cache` to add the package as a dependency
- Completes installation with `uv sync` to ensure all dependencies are properly resolved

**Client (`packages/client/Dockerfile` and `packages/client/Dockerfile.dev`)**:
- Applied same UV workflow modernization as allocator
- Uses `uv init --python 3.11 --no-readme` for consistent Python version
- Follows same pattern: init → venv → add → sync
- Changes applied to both production (`Dockerfile`) and development (`Dockerfile.dev`) variants

### 2. Additional Client Packages

Added utility packages to enhance the client VM user experience:
- **VLC**: Media playback support for video/audio files
- **zip/unzip**: Archive handling utilities for file compression/decompression
- **terminator**: Advanced terminal emulator with split-pane support and better usability than default terminal

These packages were added to both `Dockerfile` and `Dockerfile.dev` to ensure consistent functionality across environments.

### UV Installation Method
The new approach using `uv init` and `uv add` provides several benefits:
1. **Better dependency management**: UV's project-based workflow creates a proper `pyproject.toml` and lock file, ensuring reproducible builds
2. **Improved caching**: The `--no-cache` flag prevents stale cache issues during builds
3. **Consistency**: Same pattern across both allocator and client images
4. **Future-proofing**: Aligns with UV's recommended best practices for package management
5. **Explicit Python version**: Using `--python 3.11` ensures consistent Python version across deployments

### Additional Client Packages
The added packages address common user needs identified in client VM usage:
- **VLC**: Users often need to preview/playback media files during research workflows
- **zip/unzip**: Essential for handling compressed datasets and results
- **terminator**: Power users benefit from advanced terminal features like split panes

## Testing

These changes should be tested with:
1. **Docker builds**: Verify both allocator and client images build successfully
2. **Package installation**: Confirm packages are correctly installed in virtual environments
3. **Runtime verification**: Ensure console scripts and entry points work as expected

## Breaking Changes

None. These are internal implementation changes that don't affect the API or user-facing functionality.

## Future Considerations

- Monitor UV's evolution and update to newer patterns as they emerge
- Consider adding additional productivity tools based on user feedback
- May want to make client utility packages configurable via environment variables for customized deployments

### DIFF
--- a/packages/client/Dockerfile.dev
+++ b/packages/client/Dockerfile.dev
@@ -131,6 +131,9 @@ RUN mkdir -p /home/${USERNAME}/.config/google-chrome && \
 USER ${USERNAME}
 WORKDIR /home/${USERNAME}
 
+# Ensure miniforge Python is in PATH for all subsequent commands
+ENV PATH="/home/${USERNAME}/miniforge3/bin:$PATH"
+
 # Create scripts directory with miniforge, uv, and nvm installation scripts
 COPY install_miniforge.sh install_nvm.sh /scripts/
 
@@ -148,7 +151,7 @@ COPY . /home/${USERNAME}/lablink-client
 # Install client service package with dev dependencies using uv sync
 # Use UV_PROJECT_ENVIRONMENT to specify venv location (matches production)
 WORKDIR /home/${USERNAME}/lablink-client
-RUN sudo UV_PROJECT_ENVIRONMENT=/home/${USERNAME}/.venv uv sync --extra dev && \
+RUN sudo -E UV_PROJECT_ENVIRONMENT=/home/${USERNAME}/.venv uv sync --extra dev && \
     sudo chown -R ${USERNAME}:${USERNAME} /home/${USERNAME}/.venv /home/${USERNAME}/lablink-client
 WORKDIR /home/${USERNAME}
 


packages/allocator/Dockerfile
```diff
@@ -29,8 +29,10 @@ WORKDIR /app
 COPY pg_hba.conf /etc/postgresql/15/main/pg_hba.conf
 
 # Install package from PyPI using UV in explicit venv location
-RUN uv venv /app/.venv && \
-    uv pip install --python=/app/.venv/bin/python lablink-allocator-service==${PACKAGE_VERSION}
+RUN uv init --python 3.11 --no-readme && \
```
eberrigan: we can commit the python version if you prefer that, i believe that is supposed to be best practice https://docs.astral.sh/uv/concepts/python-versions/#python-version-files

Issue #50, opened 2025-05-27
Add unassign vm API so that the admin can unassign unused VM


7174Andy:
Currently, there are no ways for both the admin and the client to un-assign VM instances that are already assigned to a specific user. We can create a new API endpoint so that the admin can unassign VM instances in the `/admin/instances` where admin can view the table of the database with instances and assigned users.

PR #346, opened 2026-05-14
refactor: remove monitoring feature (schema, CLI cost branch, docs)

7174Andy:
## Summary

Mirrors the monitoring-feature removal that shipped in `lablink-template` (companion PR: https://github.com/talmolab/lablink-template/pull/50). Removes the `monitoring` config schema, the CLI cost-estimation branch that consumed it, all test references, and the user-facing documentation that described it.

- **Schema (`packages/allocator/.../structured_config.py`):** Deletes `MonitoringConfig`, `BudgetConfig`, `CloudTrailConfig`, and `ThresholdsConfig` dataclasses. Removes the `monitoring: MonitoringConfig` field from `Config`.
- **Default config:** Drops the `monitoring:` block from `packages/allocator/.../conf/config.yaml`.
- **CLI:** Removes the `if cfg.monitoring.enabled:` cost branch and the `cloudwatch`/`cloudtrail` entries from `FALLBACK_COSTS` in `commands/status.py`. Drops monitoring re-exports from `config/schema.py`.
- **Tests:** Removes the dedicated `test_monitoring_adds_cloudwatch` test, the `monitoring` entries from `mock_cfg`, `omega_config`, and `valid_config_dict` fixtures, and the now-orphaned `test_load_deeply_nested` test.
- **Docs:** Removes the "Monitoring Options" reference section from `docs/configuration.md`, the `monitoring:` sub-blocks from example YAML configs, the `monitoring.enabled` row from `docs/cost-estimation.md`, the monitoring customization bullet from `docs/cli/first-deployment.md`, and the "monitoring" mention from `docs/index.md` and `docs/cli/index.md`.

4 commits, +4/-294 lines across 13 files.

## Migration for downstream consumers

Existing user YAML configs that still contain a `monitoring:` block will continue to load without error — `load_config` silently ignores top-level keys that are not on the `Config` dataclass (see `packages/cli/src/lablink_cli/config/schema.py:49,63`). No warning is emitted. Production behavior is unchanged — the field was never required — but operators should remove the `monitoring:` block from their configs after pulling, for cleanliness.

The companion PR's migration note covers the Terraform side (operators must `terraform apply` to destroy any deployed monitoring resources before pulling that PR; this PR has no Terraform impact on its own).

## Out of scope

- Broader CloudTrail/CloudWatch operational guidance in `docs/aws-setup.md` and `docs/security.md`. Those sections cover IAM, security checklists, and app-log handling — they're independent of the removed config schema.
- Historical CHANGELOG entries.
- The 3-level nested config loader in `cli/config/schema.py:load_config` is kept (it's still useful for forward compatibility if any future config grows a 3-level structure).

## Test Plan

- [x] `uv run pytest packages/allocator/tests packages/cli/tests --ignore=packages/allocator/tests/terraform/tests/test_plan.py` is green
- [x] `uv run ruff check packages/` is clean
- [x] `uv run mkdocs build` succeeds
- [x] `git grep -nE 'MonitoringConfig|BudgetConfig|CloudTrailConfig|ThresholdsConfig|cfg\.monitoring|monitoring\.enabled'` returns no hits in src/tests/docs (excluding `docs/superpowers/` working-document directory)
- [x] Operator removes the `monitoring:` block from their local config after pulling (no error, just cleanliness)


7174Andy (commented):
## Code Review

### Summary

This PR cleanly removes the monitoring feature across schema, CLI cost estimation, tests, and docs — mirroring the companion `lablink-template` PR #50. The removal is structurally sound: schema dataclasses are gone, the cost branch is excised, and all seven affected doc pages are updated consistently. All CI checks pass (lint, tests, Docker builds, docs build). One minor inaccuracy in the PR description and two small nits are worth noting, but none block merge.

---

### Strengths

- **Thorough sweep**: Monitoring references are removed across all four layers (schema, CLI logic, tests, docs) in a single coherent PR. No dangling `cfg.monitoring` attribute accesses remain in any source file.
- **Backwards-compatible loader**: `load_config` uses `hasattr(cfg, key)` gating, so existing user YAMLs containing a `monitoring:` block will be **silently ignored** — no crash, no user action required at load time. Verified locally.
- **Test hygiene**: The deleted `test_monitoring_adds_cloudwatch` and `test_load_deeply_nested` tests are removed rather than left as dead weight, and the `mock_cfg` fixture no longer sets `cfg.monitoring.enabled = False` on a field that no longer exists.
- **CI green**: All 11 check runs pass — lint, allocator/client/CLI tests, Docker builds, and docs build.
- **Commit granularity**: The four commits (CLI cost branch → CLI schema re-exports → allocator schema → docs) are logically sequenced and independently reviewable.

---

### Issues Found

#### 🟡 Important (should address)

- **`packages/allocator/src/lablink_allocator_service/conf/config.yaml` (last line) and `packages/cli/src/lablink_cli/commands/status.py` (last line in diff context)** — Both `config.yaml` ends without a trailing newline (`\No newline at end of file` appears in the diff), which is a POSIX convention violation and will cause a ruff/editors to flag it. The pre-existing file already had this issue on the old last line (`retention_days: 90`), but the PR now leaves it on `on_error: "continue"`. Worth fixing in a follow-up or as part of this PR's final commit.

  ```yaml
  # config.yaml last line should end with \n
    on_error: "continue"  # "continue" = ignore errors, "fail" = stop VM setup on error
  ```

- **PR description inaccuracy — "unknown-key warning"**: The PR body says *"Any existing user YAML config that still contains a `monitoring:` block will trigger an **unknown-key warning**."* This is not accurate. `load_config` silently drops unknown top-level keys (`elif hasattr(cfg, key)` evaluates `False` and the key is skipped without any `warnings.warn()` or log output). Verified by running `load_config` with a `monitoring:` block against the current code — no warning is emitted. The actual behavior (silent ignore) is arguably better for UX, but the description should be corrected to avoid confusing operators who go looking for a warning that never appears.

#### 🟢 Minor / Suggestions

- **`packages/cli/tests/test_schema.py`** — The `test_load_deeply_nested` test (which exercised the 3-level YAML loader path in `load_config`) was the only test covering that code branch. The PR description acknowledges the 3-level loader is intentionally kept for future use, but the branch is now untested. Consider adding a minimal replacement test using a synthetic 3-level config fixture (e.g., `ssl.extra_options.foo`) so the loader path doesn't silently regress. Since the PR description explicitly calls this out as a deliberate trade-off, this is a nit rather than a blocker.

  ```python
  # Example replacement to keep the branch covered without monitoring:
  def test_load_3level_nesting_noop(self, tmp_path):
      """3-level keys for unknown sub-configs are silently ignored."""
      data = {"dns": {"enabled": True, "domain": "x.example.com"}}
      # dns has no 3-level sub-configs — just confirm load doesn't error
      path = tmp_path / "config.yaml"
      with open(path, "w") as f:
          yaml.dump(data, f)
      cfg = load_config(path)
      assert cfg.dns.enabled is True
  ```

- **`docs/cost-estimation.md:322`** — The `## Cost Monitoring` section header and surrounding AWS Budgets/SNS bash commands are independent of the removed config schema (they document manual AWS CLI steps, not the `monitoring.enabled` flag). These are correctly left in place per the PR's "out of scope" callout. Just confirming this is intentional and consistent with the PR description.

- **Uncommitted local changes to `setup.py` and `test_setup_full.py`**: The working tree shows modifications to `packages/cli/src/lablink_cli/commands/setup.py` and `packages/cli/tests/test_setup_full.py` that are **not part of this PR's diff**. These look like a follow-up improvement (auto-writing `dns.zone_id` back to config rather than printing a tip). Confirm these are intentionally held back for a separate PR and won't accidentally be committed onto this branch before merge.

---

### Test Coverage

- Deleted tests (`test_monitoring_adds_cloudwatch`, `test_load_deeply_nested`, `test_monitoring_cost_entries`) are appropriately removed — they tested behavior that no longer exists.
- The `test_all_costs_positive` test previously set `cfg.monitoring.enabled = True` to cover the monitoring branch; that line is now removed and the test still runs, which correctly validates the remaining cost paths.
- The 3-level YAML loader branch in `load_config` is now untested (noted above as a minor nit).

---

### Documentation / Specs

- `docs/configuration.md`: All five `monitoring:` YAML blocks (including three in full config examples) are removed. Consistent.
- `docs/cli/index.md`, `docs/index.md`, `docs/cli/first-deployment.md`, `docs/cost-estimation.md`: Updated correctly.
- `openspec/specs/`: No monitoring-specific capability spec existed, so no OpenSpec change proposal is required. The `openspec/project.md` references to "AWS CloudWatch: Monitoring and alerting (optional)" and "CloudWatch: Monitoring alarms (optional)" are generic AWS service descriptions, not config-schema references — leaving them is correct per the PR's out-of-scope callout.
- No CHANGELOG file exists in the repo (confirmed), so the PR description's note about "historical CHANGELOG entries" being out of scope is accurate.

---

### Recommendations

1. Fix the missing trailing newline on the last line of `config.yaml` — either in this PR or a quick follow-up.
2. Correct the PR description's "unknown-key warning" claim to "silently ignored" before merge, to avoid operator confusion.
3. Consider adding a minimal replacement test for the 3-level loader path (see suggestion above).
4. Confirm the `setup.py` / `test_setup_full.py` local changes are intentionally out of scope for this PR.

---

### Questions

- Is the silent-ignore behavior on unknown YAML keys the intended contract going forward? If so, should `load_config` emit a `warnings.warn()` for unknown top-level keys to help operators detect stale config fields? That would make the PR description accurate and would be a small, low-risk addition.

PR #351, opened 2026-05-17
PR B: provider abstraction seam (behavior-preserving AWS extraction)

7174Andy:
## Summary

Introduces a provider-abstraction seam in the allocator and routes today's two AWS-coupled runtime paths through it, **behavior-preserving** (observable behavior and return values unchanged). This establishes the provider contracts that later AWS-decoupling PRs build on, while keeping this PR small and reviewable.

- **New `providers/` package**
  - `protocol.py` — `ClientHandle` dataclass, `ProviderActionNotWired`, `ClientConnectivity` / `ComputeProvider` `@runtime_checkable` Protocols; re-exports the single canonical `BrowserSessionTarget` from `client_session`
  - `connectivity/allocator_proxied.py` — `AllocatorProxiedClientConnectivity`, delegates verbatim to `client_session.prepare_browser_session`
  - `aws.py` — `AWSProvider`: `recover_hosts` → `stop_start_ec2_instance` (returns a **real accumulated bool**, not unconditional True), `list_hosts` → `terraform_utils`; `provision_hosts`/`destroy_hosts` raise `ProviderActionNotWired` (Terraform `apply`/`destroy` intentionally stay inline in `main.py` until a later PR)
  - `registry.py` — `get_provider` via `lablink.providers` entry points, default `"aws"`, builtin fallback for editable installs
- **Core wiring (behavior-preserving)**
  - `main.submit_vm_details` calls `prepare_browser_session` through `provider.client_connectivity` — identical kwargs; `RotationFailed` → 503/`update_health` path unchanged
  - `AutoRebootService` failed-VM EC2 recycle routed through `provider.recover_hosts` when `can_recover_hosts`; `record_reboot` still gated on the real bool; legacy `provider=None` path byte-identical; SSH warm/cold logic untouched
  - `aws` entry point registered in `pyproject.toml`
- **Guards & scope**
  - Scoped, direct-import-only cloud-agnosticism guard keeps `protocol.py`/`registry.py` free of AWS imports; the full-core import-graph guard is deferred to a later PR — this PR does **not** by itself achieve "core has zero AWS imports" (`main.launch()/destroy()` keep inline boto3/Terraform until a later PR)

## Test plan

- [x] 20 new tests across `tests/providers/**` and `tests/test_reboot.py`
- [x] Full allocator suite: `uv run pytest tests --ignore=tests/terraform -q` → **417 passed, 13 skipped, 0 failed** (baseline was 397 passed; +20 new, 0 regressions)
- [x] `ruff check src tests` clean (CI-equivalent scope)
- [x] `import lablink_allocator_service.main` succeeds (no import-time hazard)
- [x] Two `test_api_calls.py` monkeypatch targets repointed to the new effective patch site
- [x] Reviewer note: per-PR-B coverage — every new module is ~100% covered; the repo's local `--ignore=tests/terraform` coverage (~86%) is a **pre-existing** artifact of terraform-gated route modules, not a regression (this PR slightly raised it). Real CI runs `pytest tests --cov=. --cov-fail-under=90` (no terraform ignore).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -43,11 +43,13 @@ lablink/
 │   ├── client/             # Client service (GPU health, subscription)
 │   └── cli/                # CLI tool (Typer, deploys infrastructure)
 │       └── src/lablink_cli/
-│           ├── app.py              # CLI entry point and commands
-│           ├── terraform_source.py # Downloads Terraform files from lablink-template releases
-│           ├── commands/           # Command implementations (deploy, destroy, status, logs, etc.)
-│           ├── config/             # Config schema and validation
-│           └── tui/                # Interactive TUI (wizard, logs viewer)
+│           ├── app.py                  # CLI entry point + Typer command definitions
+│           ├── api.py                  # HTTP client for the allocator service
+│           ├── terraform_source.py     # Downloads Terraform files from lablink-template releases
+│           ├── deployment_metrics.py   # CLI-local cache of deploy metrics (~/.lablink/deployments/)
+│           ├── commands/               # Command implementations (deploy, cleanup, status, logs, doctor, setup, launch, export_metrics)
+│           ├── config/                 # Config schema and validation
+│           └── tui/                    # Interactive TUI (wizard, logs viewer)
 ├── openspec/
 │   ├── project.md          # Project conventions
 │   ├── specs/              # Capability specifications
--- a/packages/allocator/pyproject.toml
+++ b/packages/allocator/pyproject.toml
@@ -36,6 +36,9 @@ lablink-allocator = "lablink_allocator_service.main:main"
 generate-init-sql = "lablink_allocator_service.generate_init_sql:main"
 lablink-validate-config = "lablink_allocator_service.validate_config:main"
 
+[project.entry-points."lablink.providers"]
+aws = "lablink_allocator_service.providers.aws:AWSProvider"
+
 [tool.setuptools.package-data]
 lablink_allocator_service = ["terraform/**/*", "conf/*.yaml"]
 
--- a/packages/allocator/src/lablink_allocator_service/main.py
+++ b/packages/allocator/src/lablink_allocator_service/main.py
@@ -43,14 +43,12 @@ from lablink_allocator_service.utils.terraform_utils import (
 )
 from lablink_allocator_service.scheduler import ScheduledDestructionService
 from lablink_allocator_service.reboot import AutoRebootService
-from lablink_allocator_service.client_session import (
-    prepare_browser_session,
-    RotationFailed,
-)
+from lablink_allocator_service.client_session import RotationFailed
 from lablink_allocator_service.signed_cookie import (
     sign,
     get_or_create_cookie_secret,
 )
+from lablink_allocator_service.providers.registry import DEFAULT_PROVIDER, get_provider
 from lablink_allocator_service.routes.desktop import bp as desktop_bp
 from lablink_allocator_service.routes.internal_proxy_auth import (
     bp as internal_proxy_auth_bp,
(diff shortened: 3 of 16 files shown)


7174Andy (commented):
## Code Review

### Summary
Introduces a `providers/` package (`protocol`, `registry`, `aws`, `connectivity/allocator_proxied`) and routes two AWS-coupled runtime paths through it: per-session browser prep in `submit_vm_details` and the EC2 stop/start fallback in `AutoRebootService._reboot_vm`. Terraform `apply`/`destroy` deliberately stay inline. Scope is well-contained, the abstraction boundary is well-placed (protocol/registry stay AWS-free), and the PR description is unusually honest about what it does *not* achieve. **Overall: approve-quality with minor non-blocking nits.**

### Strengths
- Behavior preservation verified by tracing equivalence, not just trusting the claim (see below).
- Strong test coverage of the seam (20 tests): protocol runtime-checkability, registry default/unknown/entry-point, AWS recover bool accumulation incl. partial-failure → `False`, connectivity delegation, both new reboot branches, scoped import guard.
- `test_sr_f1_imports.py` is honestly scoped and documents that it does *not* follow the transitive `protocol → client_session → aws_utils` path; deferring full import-graph enforcement is the right call for an incremental PR.
- Repointed `test_api_calls.py` monkeypatch targets correctly move to the new effective patch site.

### Behavior-preservation verification
- **Legacy `provider=None`** (`reboot.py:278-281`): byte-identical `stop_start_ec2_instance(instance_id, region=self.region)`. ✓
- **Production path**: `reboot_service` is constructed with `provider=app.config["LABLINK_PROVIDER"]`, always a wired `AWSProvider` (`can_recover_hosts=True`). So production now takes the **new** `provider.recover_hosts` branch, not the legacy `else`. Equivalence holds: one `ClientHandle` → `provider_metadata.get("region", …)` resolves to `self.region` → same `stop_start_ec2_instance(instance_id, region=self.region)`, bool propagated to the unchanged `if success: record_reboot` gate. Functionally identical. ✓
- `submit_vm_details` connectivity delegation is verbatim kwargs pass-through; `RotationFailed → 503/update_health` path untouched (`main.py:309-324`). ✓

Worth stating plainly: "behavior-preserving" rests on an *equivalence argument* + tests, not on production running literally unchanged code. The argument checks out.

### Issues Found

#### Critical
- None.

#### Important
- None blocking.

#### Minor / Suggestions
- **Dead defensive fallback** (`main.py:299-301`): `app.config.get("LABLINK_PROVIDER") or get_provider(None, …)` — `LABLINK_PROVIDER` is set unconditionally at module load (`main.py:68`), so the `or get_provider(...)` is unreachable; if reached it would re-run `entry_points()` discovery per request. Recommend just `app.config["LABLINK_PROVIDER"]`. This is the only thing I'd change before merge.
- **Local import could be hoisted** (`reboot.py:263`): `from …providers.protocol import ClientHandle` inside `_reboot_vm`. No circular-import reason exists (the `providers` graph never imports `reboot`); move to module top for consistency.
- **Naming collision risk for later PRs** (`main.py:69`): `getattr(cfg, "provider", None)` resolves to `None` today (no top-level key; relies on OmegaConf raising an `AttributeError` subclass under struct mode — which it does, so not a bug). But the schema already has `cfg.ssl.provider`. A future top-level compute `cfg.provider` will sit confusingly next to it — consider `cfg.compute_provider`/nested key when that config is actually introduced. Not actionable here.

#### Informational
- `ClientConnectivity.prepare_browser_session(self, **kwargs)` (`protocol.py:33`) drops the explicit kwarg contract — appropriate for a thin seam, but a caller typo now fails at runtime vs. type-check. Reasonable follow-up to tighten once stabilized.
- `provision_hosts`/`destroy_hosts` advertise capability (`can_provision_hosts=True`) but raise `ProviderActionNotWired`. Safe and tested in this PR; just a latent footgun if a future caller trusts the flag without handling the raise.

### Testing & Conventions
- The unchecked coverage box in the description is consistent with the established understanding that the 90% gate is CI-only (full `tests`, no terraform ignore) and local `--ignore=tests/terraform` ~86% is a pre-existing artifact, not a regression — no concern.
- Note: the "417 passed / 0 regressions" claim was not independently re-run; the wiring/equivalence above was verified by reading.

### Security
No new secrets or injection surface. `provider` name from config is validated against a known dict (`ValueError` on unknown — no arbitrary import from config); entry-point `ep.load()` resolves only from installed package metadata. Fine.

### Recommendations
1. Simplify `main.py:299-301` to `app.config["LABLINK_PROVIDER"]` (drop unreachable fallback).
2. Hoist the `ClientHandle` import in `reboot.py` to module top.
3. (Later PRs) Reconsider top-level compute `provider` config key naming vs. existing `cfg.ssl.provider`.

PR #352, opened 2026-05-18
PR C: registration API + per-client credentials (AWS-only)

7174Andy:
## Summary

Phase 2 / **PR C** of the AWS-decoupling roadmap (successor to #351 / PR B's provider seam; precedes PR D). AWS-only and behavior-preserving for provisioning.

- **`POST /api/v1/clients/register`** — validates the deployment register-token (argon2-hashed at rest in `settings`, "Option A"), mints a per-client `client_secret` returned **once**, idempotent on `machine_identity` (adopt the terraform-pre-created hostname row → re-register/rotate → insert).
- **`GET /api/v1/clients/<id>/status`** — client-facing versioned status, per-client-secret auth. Supersedes the now-removed `GET /api/vm-status/<host>`.
- **Auth swap** — `heartbeat`, `POST /api/vm-status`, `gpu_health`, `update_inuse_status` move from the deployment-wide `API_TOKEN` to per-client `client_secret` via a new `require_client_secret` decorator. Log shipping (`/api/vm-logs`) and `/api/vm-metrics` intentionally stay on `API_TOKEN`.
- **Schema** — Phase 2 columns/indexes (`machine_identity`, `provider`, `endpoint_url`, `provider_metadata`, `client_secret_hash`, `gpu_*`, `last_release_time`) added directly to `generate_init_sql.py` (one-time-use DB → no runtime migration).
- **Provider seam** — first consumer of PR B's contract: `ClientJoinMaterial` + `make_join_material` on `ClientConnectivity`, AWS impl.
- **Cloud-init / client** — `user_data.sh` registers at boot **before** the container, threads `CLIENT_SECRET` in; client `http_utils` and `start.sh send_status` prefer `CLIENT_SECRET` with `API_TOKEN` fallback.
- New `secret_hash.py` (argon2); `argon2-cffi` dependency added.

**Deferred to PR D:** ManualProvider, the `lablink register` CLI, the wizard, register-token keychain distribution, `admin_password_hash`. No `lablink-template` cross-repo change.

## Atomicity fix (`a780c42e`)

The initial `register_client()` was a check-then-act across multiple statements; under the pool's `ISOLATION_LEVEL_AUTOCOMMIT` connections that is **not** atomic, so concurrent/overlapping registers could (a) return 200 with a `client_secret` whose hash was never persisted (silent permanent auth failure), or (b) surface the documented 409 race as an unhandled 500.

Replaced with a **single atomic statement**:

```sql
INSERT INTO <vm> (...) VALUES (...)
ON CONFLICT (hostname) DO UPDATE SET ...
WHERE <vm>.machine_identity IS NULL
   OR <vm>.machine_identity = EXCLUDED.machine_identity
RETURNING hostname;
```

- Covers fresh INSERT / adopt (`machine_identity` NULL) / rotate (same `machine_identity`) in one round-trip.
- **No-hijack guard**: a hostname owned by a *different* `machine_identity` → `WHERE` false → `RETURNING` empty → `None` → route returns **409**.
- `machine_identity` already bound to another hostname → partial-unique index → `psycopg2.IntegrityError` → route returns **409**.
- `routes/registration.py` now wires both 409 paths (`None` and `IntegrityError`); a test asserts `execute.call_count == 1` so it can't regress to multi-step.

## Dead-code removal (bundled cleanup)

Verified zero callers (client pkg, CLI, terraform/cloud-init, dashboard templates, internal code):

- `POST /vm_startup` — vestigial CRD-era boot ping, subsumed by registration + heartbeat + `POST /api/vm-status` (`cb158251`).
- `POST /api/reboot-vm` + `GET /api/reboot-vm/<hostname>` and `tests/test_reboot_api.py` — failed-VM recovery actually runs in-process via `AutoRebootService`; these HTTP routes were never invoked (`a9611f17`).
- `GET /api/vm-status/<hostname>` — superseded by `GET /api/v1/clients/<id>/status` (`042efbc6`).
- `database.get_reboot_info` — orphaned after the reboot-vm routes were removed (`6fb3ecca`).

In-process recovery (`AutoRebootService`, scheduler, `record_reboot`, `ensure_reboot_columns`, `get_failed_vms`, `release_assignment`, `tests/test_reboot.py`) is unchanged. `chore: remove unnecessary comments` (`e246d49a`) is comment/docstring-only.

## Known follow-ups (non-blocking)

- **argon2 on the heartbeat hot path** — `require_client_secret` runs `verify_secret` (default `PasswordHasher`, ~64 MiB/call) per request on `/api/heartbeat` etc. Acceptable at tutorial scale; revisit (lower params or short verification cache) before high-VM-count deployments.
- **Contract change** — identity-less requests to the swapped endpoints now return `401 "client identity required"` (decorator) instead of the old `400 "Hostname is required."`. M2M-only; noted here for completeness.
- **Minor:** not-found hostname returns 401 without an argon2 verify (timing oracle for hostname enumeration; hostnames are predictable anyway); the register response echoes `register_token` (unused by `user_data.sh`); `database.get_vm_by_machine_identity()` is now unused (the atomic rewrite dropped its only caller) — removal candidate for PR D; genuinely-concurrent re-register of the *same* VM is last-writer-wins (DB stays consistent — this is no longer the pre-fix silent corruption).

## Test plan

- [x] Allocator suite, terraform-ignored: **430 passed / 13 skipped** (deleted: 4 vm_startup + reboot-api + 3 reboot + 3 vm-status/<host> tests; added: register/secret_hash/atomicity/409). CI runs the full suite incl. `tests/terraform` for the 90% gate.
- [x] Client suite: **67 passed**. `ruff check src tests` (allocator + client): clean.
- [x] Atomicity: single-statement `register_client` (asserts one `execute`); `None` → 409; `IntegrityError` → 409; no-hijack `WHERE` guard present.
- [x] End-to-end AWS boot: register → `client_secret` minted → container heartbeats with it → readiness poll observes `running`; wrong/missing/rotated register-token → 401; idempotent re-register doesn't duplicate.
- [x] Legacy `/api/vm-logs` and `/api/vm-metrics` still authenticate with `API_TOKEN` (removed: the dead legacy `GET /api/vm-status/<host>`).


7174Andy (commented):
## Code Review — PR C: registration API + per-client credentials

Solid, well-tested PR: parameterized SQL throughout, argon2 at rest, `client_secret` returned exactly once, dual-auth boundary preserved (log shipping + `/api/vm-metrics` correctly stay on `API_TOKEN`), schema goes straight into `generate_init_sql.py` (correct for the one-time-use DB model), and the auth swap is consistent across every client caller via `get_client_env()` with an `API_TOKEN` fallback. The parametrized negative-auth tests (`CLIENT_SECRET_PROTECTED_ENDPOINTS` × no-token / wrong-secret / non-bearer) are excellent.

One critical correctness issue, plus a few important items.

### Critical

**1. `register_client()` is not atomic — connections run in `ISOLATION_LEVEL_AUTOCOMMIT`**

`_PooledCursor.__enter__` forces `ISOLATION_LEVEL_AUTOCOMMIT` on every checkout (`database.py:38`), so each `cursor.execute()` inside `register_client` (`database.py:296-367`) is its own committed transaction. The SELECT-then-UPDATE/INSERT is a check-then-act with no transaction or row lock. Two failure modes under concurrent / repeated registration:

- **Adopt branch (silent auth failure, no error raised):** two requests both miss the `machine_identity` lookup, both find the same `machine_identity IS NULL` hostname row. First `UPDATE ... WHERE hostname=%s AND machine_identity IS NULL` commits. The second matches **0 rows** — but the code still `return hostname` (`database.py:354`) and the endpoint returns 200 with a freshly minted `client_secret` whose hash was never persisted. That VM then fails every authenticated call with 401 forever, with no error anywhere.
- **Insert branch:** two requests both miss both lookups, both `INSERT`; the partial-unique index raises `UniqueViolation` on the second. The comment at `database.py:356-358` explicitly says "the endpoint maps that to a retry-safe 409 (per spec)" — but `routes/registration.py:40` calls `register_client()` with **no try/except**, so it surfaces as an unhandled 500, not a 409.

This isn't purely theoretical: memory note — the lablink-client base AMI re-runs `user_data` on *every* boot (scripts-user `frequency=always`), so a re-register overlapping an in-flight one, or a cloud-init retry, is a realistic path, and idempotent registration is the core guarantee of this PR.

Recommend: make the whole operation atomic — either run `register_client` in a single non-autocommit transaction with `SELECT ... FOR UPDATE` on the candidate row, or collapse it to an `INSERT ... ON CONFLICT (machine_identity) ...` upsert. And implement the 409 path in `routes/registration.py` (catch `psycopg2.IntegrityError` → 409) so it matches the in-code comment, or delete the comment if 409 isn't actually wired.

### Important

**2. Missing 409 handler contradicts documented behavior** — see above; `database.py:356` documents a 409 backstop that does not exist in `registration.py`. At minimum the comment is misleading.

**3. Per-request argon2 on the heartbeat hot path** — `require_client_secret` runs `verify_secret` (default `PasswordHasher()`: ~64 MiB + tens of ms per call) on every `/api/heartbeat`, `/vm_startup`, `/api/vm-status`, etc. Heartbeats are periodic per VM, so this is a standing CPU/64-MiB-alloc cost per heartbeat, and anyone who knows a (predictable) hostname can force argon2 work. Probably acceptable at tutorial scale, but please acknowledge the load characteristic — consider lower argon2 params for these high-frequency M2M tokens or a short-lived in-memory verification cache.

**4. Undocumented API behavior change** — endpoints that previously returned `400 {"error": "Hostname is required."}` now return `401 {"error": "client identity required."}` because `require_client_secret` rejects identity-less requests before the view runs (visible in the `test_api_calls.py` assertion changes). Reasonable (can't auth without identity), but it's a contract change for these M2M endpoints and isn't called out in the PR description.

### Minor / Suggestions

- **Hostname enumeration timing oracle:** unknown hostname → fast 401 (no argon2); known hostname + wrong secret → slow argon2. Hostnames are predictable (`{prefix}-vm-{n}`) so impact is low, but a dummy `verify_secret` on the not-found path would close it.
- **`register_token` echoed back in the response** (`registration.py`, via `ClientJoinMaterial`): `user_data.sh` only reads `client_secret`; echoing the register-token into the JSON body just adds it to logs. Consider scoping it to the manual flow (PR D) rather than the AWS response.
- **`get_vm_by_machine_identity()` (`database.py:274`) is added + tested but unused** in this PR — confirm it's intended for PR D, otherwise it's dead code.
- **`last_release_time` column + `_assignable_idx`** are added but unused until PR D — fine and roadmap-aligned, just noting the schema lands ahead of its consumer.
- **`user_data.sh` readiness poll uses `$VM_NAME`** rather than the returned `client_id`. Consistent today (adopt branch returns `VM_NAME`), but it'll silently poll the wrong path if the rotate branch ever returns a stored hostname ≠ `VM_NAME`. Prefer polling the returned `client_id`.
- Terraform escaping in `send_status` (`$${CLIENT_SECRET}` vs single-`$` `$CLIENT_SECRET` elsewhere) is correct but subtle — worth a one-line comment so a future edit doesn't break it.

### Question

The register-token is regenerated each allocator process start (`REGISTER_TOKEN = secrets.token_urlsafe(32)`) and re-hashed into `settings` on every `init_database()`. If the allocator restarts after VMs launch but before they register, their `user_data` register-token is stale and registration fails. This mirrors the existing `API_TOKEN` model — confirming this is intended/acceptable for the deployment model rather than an oversight.

Overall: strong implementation and test coverage; the registration atomicity issue (#1) is the one that should block merge until the check-then-act is made transactional and the 409 path either implemented or the comment corrected.

PR #357, opened 2026-05-22
PR D3: operator-side manual mode (configure wizard + compose deploy)

7174Andy:
## Summary

Closes the manual-BYO loop on the operator side: configure the allocator with `provider: manual`, deploy it via docker-compose, and manage it with the existing CLI commands. Each affected command dispatches on `cfg.provider`, so the AWS path is preserved bit-for-bit.

Initial staging surfaced three follow-on issues that this branch also addresses:
- **D1's LAN-direct viewer never authenticated** — KasmVNC's HTTP BasicAuth is browser-unreachable on a WebSocket upgrade in modern Chromium/Firefox. Pivoted to RFB-protocol **VncAuth**, then peeled two more layers (DES bit-ordering, KasmVNC fork clobbering the credential in JS) before the handshake actually succeeded. See [design](#lan-direct-browser-auth-pivot-basicauth--vncauth) below.
- **`lablink register` never re-pulled the client image** — a republished tag would silently ship stale bits to the BYO box. Added `--pull always` to the `docker run` argv.
- **A handful of compose / CLI / allocator nits** (BYO port publishing, manual-provider Terraform skip, reboot path, GHCR image names, etc.).

**Predecessors merged:** PR A (#344) · PR B (#351) · PR C (#352) · PR D1 (#353) · PR D2 (#354).

After this PR the full participant workflow is runnable without any hand-curl or `docker run` paste: operator runs `lablink configure → deploy`, BYO box owner runs `lablink register`, student opens the allocator URL.

### Allocator (foundation)

- `Config` dataclass gains `provider: str = field(default="aws")`, validated against `("aws", "manual")` in `get_config_errors`.
- `main.py` reads `cfg.provider` instead of the hardcoded `DEFAULT_PROVIDER` (both at module load and in the `/api/request_vm` fallback).
- Gated `REGISTER_TOKEN=%s` log line at startup (only when the configured provider's `can_provision_hosts` is False — capability-flag gating sidesteps the cleanliness-guard ban on `provider == "..."` literals in `main.py`). The format is the cross-process contract that the CLI compose deploy uses to surface the token from `docker logs`.

### CLI wizard

- **`ProviderScreen`** inserted between Deployment Identity and Region: picks AWS vs Manual; force-resets `ssl.provider` and `machine.image` to manual defaults when switching to manual.
- **`ManualMachineScreen`**: single-input step for the BYO client image. Manual flow skips Region / Machine instance-type / EIP.
- **`DnsScreen`** refactored from index-based to id-based `RadioSet` dispatch (each option now has a stable `dns-<provider>` id). Adds a 5th **self-signed** option, conditionally hidden when public-TLS options are shown. `_next` branches to `ReviewScreen` for manual (skipping `StartupScreen`).
- Step labels normalized: dropped `of 6` suffix everywhere (manual is 5 screens, AWS is 7 — a single count would mislead).

### Operator workflow commands

Each gates manual at the top of `run_*`, then dispatches to a `_run_*_manual` helper. AWS bodies are byte-identical.

| Command | Manual behavior |
|---|---|
| `configure` | Skips auto-`setup`; prints "run `lablink deploy` next." |
| `setup` | No-op + friendly message (no S3 / DynamoDB needed). |
| `deploy` | Dispatches to `run_deploy_compose`. |
| `destroy` | Dispatches to `run_destroy_compose`; gains `--purge` flag. Requires typing `yes` to confirm, matching AWS destroy pattern. |
| `status` | `docker compose ps` + `/api/health` check. |
| `launch-client` | Friendly no-op (exit 0). |
| `logs` | Tails `docker logs -f lablink-allocator`. |
| `cleanup` | `docker compose down --volumes` + removes compose working dir. |
| `doctor` | Checks `docker` + `docker compose version`; keeps AWS terraform/aws checks for AWS configs. |

### New module — `commands/deploy_compose.py`

- **Single-service** `docker-compose.yml` template (packaged via `pyproject.toml`). The allocator image is monolithic — it bundles its own Postgres internally, so the stack is one container, not two. Mounts `./config.yaml` at `/config/config.yaml`; persists internal Postgres state on a named volume at `/var/lib/postgresql`. Pins `container_name: lablink-allocator` so the CLI helpers can address it.
- `render_compose_dir(cfg, target)`: copies the bundled template, writes `.env` with only `ALLOCATOR_IMAGE` / `HTTP_PORT` / `HTTPS_PORT`.
- `run_deploy_compose(cfg, *, yes, workdir_root)`: preflight (SSL must be `none` or `self_signed`; `docker` on PATH) → `resolve_admin_credentials` → confirm → render → `docker compose up -d` → health-poll → print summary with the `register_token` extracted from `docker logs lablink-allocator`.
- `run_destroy_compose(cfg, *, yes, purge, workdir_root)`: `docker compose down` (preserves data); `--purge` adds `--volumes` and `shutil.rmtree(workdir)`.

## Post-staging fixes (commits since the initial D3 push)

| Commit | What it fixes |
|---|---|
| `33d53d3` allocator: skip terraform init when provider can't provision hosts | `ManualProvider.can_provision_hosts == False` ⇒ skip the unconditional `terraform init` that would otherwise complain about a missing AWS backend. |
| `16febd5` CLI: fix manual-provider compose deploy reachability and token discovery | `lablink deploy` for manual now health-checks the compose stack on `127.0.0.1` (not the operator's WAN IP), and `register_token` extraction tolerates the new line format. |
| `d408e4a` allocator: skip AWS reboot path for manual/BYO providers | The cron-style reboot scheduler unconditionally called `provider.recover_hosts`; for `ManualProvider` that's a `NotImplementedError`. Gate on `can_recover_hosts`. |
| `0424216` CLI: publish BYO client ports explicitly instead of `--network host` | `docker run … -p 6080:6080 -p 7070:7070` instead of `--network host`. |
| `ddd85f4` CLI + allocator: add `lablink client unregister` and `client` subgroup | New `DELETE /api/v1/clients/<hostname>` + CLI plumbing. |
| `9b3343a` CLI: pin allocator compose service to linux/amd64 | Forces the right arch on Apple-silicon operators running the compose stack locally. |
| `9c06b4d` CLI: fix wrong GHCR image names in manual-provider deploy | Manual-provider compose template now points at the actual published image tags. |
| `e7a93df` CLI: make command help/docstrings provider-aware | `--help` text reflects what each command actually does in the current provider mode. |
| `a23b854` CLI: `lablink register` uses `docker run --pull always` | Default `--pull missing` reuses any locally-cached layer for the same tag, so republished images silently ship stale bits to the BYO box. `--pull always` does a manifest HEAD per register; only changed layers are downloaded. |

## LAN-direct browser-auth pivot (BasicAuth → VncAuth)

This was the largest sub-thread; spans four commits and three distinct root causes. **All resolved and verified end-to-end** on a real BYO same-LAN deployment (Mac allocator ↔ Windows laptop client, both on 192.168.1.0/24) — RFB `SecurityResult: 00000000`, `kasmvnc.log` shows `Connections: closed: ... (Clean disconnection)` instead of `AuthFailureException`.

### Origin

**Surfaced by:** first real-network Layer-3 smoke against PR D1's lan_direct path. `kasmvnc.log`: `Authentication attempt failed, BasicAuth required, but client didn't send any`.

**Root cause:** D1 spec'd that the bundled noVNC would attach `Authorization: Basic base64('kasm_user:<pw>')` on the WS upgrade. Modern Chromium/Firefox strip `user:pass@` userinfo from `new WebSocket("ws://…")` URLs at the URL-parser level (verified empirically — `kasmvnc.log` shows **zero TCP attempts** when userinfo is present), and the JS WebSocket API has no way to attach an arbitrary `Authorization` header to a WS upgrade. `lan_direct` has no allocator proxy in the byte path, so no server-side header injection either — the credential pipeline was unimplementable as specified.

### The pivot

**Drop HTTP-Basic on the client in `lan_direct` mode; use RFB-protocol VncAuth instead.** The bundled KasmVNC v1.4 noVNC reads `?password=` from the URL and drives the password through the in-band VNC auth handshake after the WS upgrade — no HTTP headers involved. AWS path is byte-identical (gated on the client container's existing `CONNECTIVITY` env var).

| File | Change |
|---|---|
| `packages/client/start.sh` | Branch on `CONNECTIVITY`. `lan_direct` ⇒ seed `~/.vnc/passwd` (RFB format) and launch `Xvnc` with `-DisableBasicAuth -SecurityTypes VncAuth -PasswordFile …`. Else: keep `-SecurityTypes None -KasmPasswordFile …`. |
| `packages/client/src/lablink_client_service/agent/kasmvnc.py` | `rotate_kasmvnc_password` dispatches on `CONNECTIVITY`. New `_vncauth_blob()` writes the 8-byte RFB password file. |
| `packages/client/pyproject.toml` | Adds `cryptography>=42,<46`. |
| `packages/allocator/src/lablink_allocator_service/providers/connectivity/lan_direct.py` | Mints `secrets.token_urlsafe(6)` ⇒ exactly 8 chars. VncAuth truncates longer values silently. |
| `packages/allocator/src/lablink_allocator_service/routes/desktop.py` | LAN-direct branch redirects via `location.replace` to `/static/novnc/vnc.html?host=…&port=…&password=<urlencoded>`. AWS branch is regression-locked by `test_desktop_aws_byte_identical`. |
| `packages/allocator/Dockerfile{,.dev}` *and* `packages/client/Dockerfile{,.dev}` | `sed`-patch the bundled noVNC; see "Two more bugs" below. |

### Two more bugs after the pivot, both surfaced by real-network smoke

The initial pivot commit (`126b96f5`) shipped logically-correct code that still failed end-to-end. Two latent issues, fixed in `2c460bc9` and `f884edb4`:

**(a) d3des reads key bytes in LSB-first bit order.** Standard MSB-first DES libraries (`cryptography`, `openssl enc -des-ecb`) read them MSB-first. The first `_vncauth_blob` drove `cryptography` with the raw `0x17, 0x52, 0x6B, …` fixed key — KasmVNC then silently decrypted the file to the *wrong* 8-byte plaintext, and every `VncAuth(2)` request failed with `SConnection: AuthFailureException`. **Cross-verified vs `tigervncpasswd` on Ubuntu 22.04**: password `"abcd1234"` must encrypt to `237bbe0803430cc2`, not `43252e8205851e4e`. Fix: bit-reverse the fixed-key bytes once at module load and use that as the DES key. The password as plaintext is **not** bit-reversed — only the key — which is the asymmetry that makes the round-trip work. Test vector locked.

**(b) KasmVNC's bundled noVNC fork hardcodes `this._rfbCredentials.password=""` immediately before `genDES`** in `_negotiateStdVNCAuth`. So even with the file format right, the page reads `?password=<cred>` correctly, then the JS overwrites it to empty bytes *just before* encrypting the challenge — KasmVNC compares the empty-key DES response against its expected response and rejects with `AuthFailureException` (no useful log line beyond that). Presumably intentional for Kasm's normal HTTP-BasicAuth product flow, but lethal in lan_direct. Fix: `sed -i 's/this\._rfbCredentials\.password="";//g'` against `/usr/share/kasmvnc/www/assets/ui-*.js` at image build time. Targeted (one minified statement, no other matches in the bundle) and idempotent on rebuild. **Important:** the bundle the **browser** loads is served by the **allocator's** nginx from the **allocator** container — see `lablink-nginx.conf:43-46`. The first pass put the sed only in the client Dockerfile, where it had no effect on the browser; `f884edb4` adds the same sed to `packages/allocator/Dockerfile{,.dev}`.

### Tradeoffs documented in `docs/superpowers/specs/2026-05-27-lan-direct-vncauth-pivot-design.md`

- VncAuth caps at 8 bytes / ~48 bits per session; per-session rotation + KasmVNC's `brute_force_protection.blacklist_threshold: 5` bounds online guessing.
- DES is "weak crypto"; the per-session-rotated ciphertext at rest mitigates this — the file is recoverable only with knowledge of the same DES fixed-key VNC uses, and the password rotates each request.
- The credential lands in the WS-viewer URL string (DevTools visibility). `location.replace` keeps it out of session history; `access_log off` on `/static/novnc/` keeps it out of nginx logs. Same exposure surface as any browser-direct auth scheme.

### Rejected alternatives

- **Allocator-proxy for `manual`** (what AWS uses) — BYO topology doesn't allow it; verified empirically that a co-resident allocator container can't reach the BYO host's published `:6080`.
- **WebSocket monkey-patch + userinfo URL** — Chromium strips userinfo before any TCP attempt. `kasmvnc.log` showed **no** new connection lines after the call, confirming.
- **Sidecar reverse-proxy on the client** (Sec-WebSocket-Protocol or `?token=`) — kept as a future hardening pass; overkill for D3 given lan_direct + VncAuth works.
- **Patching/forking the KasmVNC noVNC source proper** — Dockerfile `sed` is targeted enough and survives content-hash bumps via the `ui-*.js` glob; full vendoring would be a much bigger maintenance load.

## Architectural decisions worth knowing

- **Single-service compose, not two-service.** The allocator container already bundles Postgres. Splitting it would either conflict with the internal one or require a slimmed-down image variant. Single-service matches AWS deployment exactly.
- **Admin/DB creds in `config.yaml`, not env vars.** The monolithic allocator reads from `/config/config.yaml` (`CONFIG_DIR=/config` baked into the Dockerfile).
- **`REGISTER_TOKEN` exposed via log line, not Terraform output.** Manual path doesn't have Terraform, so the allocator logs `REGISTER_TOKEN=<value>` at startup. CLI extracts it via regex from `docker logs`.
- **`CONNECTIVITY` env var on the client container is now load-bearing for auth mode.** `lan_direct` → VncAuth + DisableBasicAuth + RFB password file. Anything else → BasicAuth flow.

## Out of scope / deferred follow-ups

- Public TLS providers (`letsencrypt`/`acm`/`cloudflare`) for manual deploy.
- Cross-LAN connectivity (Tailscale/TURN) — still a D1 follow-up.
- `GET /api/v1/clients` inventory endpoint.
- `quick-register.sh` curl bootstrap one-liner.
- Migrating the compose template to `lablink-template` repo (parity with Terraform configs).
- VeNCrypt `Plain` SecurityType (no 8-char limit) as a future lan_direct hardening — kept VncAuth here because the bundled noVNC's `?password=` URL hookup goes through VncAuth.

## Backwards compatibility

- Existing `config.yaml` files without a `provider:` line load identically — the field defaults to `"aws"`.
- All existing AWS `deploy`/`destroy`/`status`/`logs`/etc. paths are byte-identical to pre-D3.
- AWS client containers (without `-e CONNECTIVITY=lan_direct`) continue to use the BasicAuth + nginx auth_request flow unchanged. `test_desktop_aws_byte_identical` regression-locks the AWS redirect string. The Dockerfile sed touches only `_negotiateStdVNCAuth`, which the AWS flow never enters (BasicAuth at WS upgrade gates the connection before RFB security negotiation begins).

## Tests

| Suite | Status |
|---|---|
| Allocator (`--ignore=tests/terraform`) | 475 passed / 14 skipped |
| Client | 71 passed |
| CLI | 365 passed / 1 deselected |
| New tests | `test_vncauth_blob_matches_tigervnc_vncpasswd` (locks the DES output to `237bbe0803430cc2` for `"abcd1234"`, cross-verified vs `tigervncpasswd` and `openssl enc -des-ecb -K 17526b06234e5807 -nopad -provider legacy`), `test_vncauth_blob_truncates_to_8_chars`, `test_rotate_lan_direct_writes_vncauth_file`, `test_desktop_lan_direct_renders_direct_with_credential`, `test_desktop_lan_direct_urlencodes_credential`, register `--pull always` argv assertion. |

`ruff check packages/cli/src packages/cli/tests packages/allocator/src packages/allocator/tests packages/client/src packages/client/tests` clean. `ruff format` intentionally not run (per project convention on scoped changes).

## Test plan

- [x] Allocator unit tests pass: `cd packages/allocator && PYTHONPATH=src pytest --ignore=tests/terraform`
- [x] Client unit tests pass: `cd packages/client && PYTHONPATH=src pytest`
- [x] CLI unit tests pass: `cd packages/cli && PYTHONPATH=src pytest`
- [x] CI runs the full allocator suite with the 90% coverage gate (terraform-included)
- [x] Lint clean across all four packages
- [x] AWS path regression: `test_desktop_aws_byte_identical` still passes byte-identically
- [x] **Manual path end-to-end smoke** (Mac allocator + Windows-laptop BYO client on the same /24): registered, requested a VM, browser opened `/desktop`, RFB VncAuth handshake completed with `SecurityResult: 00000000`, `kasmvnc.log` shows `Connections: closed: ... (Clean disconnection)` and **no `AuthFailureException`** line.
- [x] `lablink deploy --help` mentions docker-compose; `lablink destroy --help` shows `--purge`
- [x] Existing config.yaml files (pre-D3) without a `provider:` line still load and behave as AWS


7174Andy (commented):
## Code Review — PR #357 (D3 manual mode)

### Overview

This is a big surface (55 files, +3562/-187) but the design is well-scoped: the AWS path is preserved bit-for-bit and the manual/BYO path is a parallel branch dispatched by `cfg.provider`. The capability-flag pattern (`can_provision_hosts`, `can_recover_hosts`) is consistently applied as the gating mechanism instead of `provider == \"...\"` literals. The PR description thoroughly documents the three latent issues that came up during staging and the reasoning behind each fix.

I reviewed the full diff and read the load-bearing files in full (`kasmvnc.py`, `desktop.py`, `lan_direct.py`, `deploy_compose.py`, `start.sh`, the wizard refactor). Verdict: **LGTM with a couple of small recommendations below — none blocking.**

### Strengths

- **Capability-flag gating is the right abstraction.** `main.py:241`, `reboot.py:220`, and the `terraform init` skip all branch on `can_provision_hosts` / `can_recover_hosts` rather than provider-type equality. The test at `tests/test_reboot.py:1064` explicitly proves the manual provider makes **zero** AWS API calls (no `NoCredentialsError` loop), which is a great regression guard for the bug the change addresses.
- **VncAuth crypto is meticulous.** `kasmvnc.py:_vncauth_blob` documents the bit-reversal asymmetry (key bit-reversed, plaintext not), the 3DES-EDE → DES collapse trick, and the atomic-write pattern. The test vector `\"abcd1234\" → 237bbe0803430cc2` is locked against `tigervncpasswd` (`test_vncauth_blob_matches_tigervnc_vncpasswd`).
- **Dockerfile sed uses verify-patch-verify.** If a future KasmVNC bundle refactors `_negotiateStdVNCAuth`, the build fails loud instead of silently producing broken images. The same patch is applied to the allocator Dockerfile too — important because the browser loads the noVNC bundle from the **allocator's** nginx, not the client's.
- **Enumeration-resistant unregister.** `routes/registration.py:152` returns 401 (not 404) when the `client_id` has no stored hash, preventing callers from probing registered IDs. Tested at `test_unregister_rejects_unknown_client_id`.
- **Token extractor regression-guards the stderr footgun.** `_extract_register_token` merges stderr into stdout via `subprocess.STDOUT`, and `test_merges_stderr_into_stdout` explicitly asserts the kwarg — same root cause as the bug the recovery-hint message warns about (`docker logs … 2>&1 | grep`).
- **AWS path byte-identical claim is regression-locked.** The proxied branch in `desktop.py` is the first early-return, and the existing test (`test_desktop_aws_byte_identical`) was not modified.
- **`.env` mode 0600**, admin/DB creds in mounted config.yaml (not env vars), `_compose_up` summary prefers LAN IP over localhost for the BYO-registration hint with a clear warning when only loopback is available — all small things done right.

### Concerns / Suggestions

#### 1. (Low) Breaking CLI rename without alias or CHANGELOG entry

`lablink launch-client` → `lablink client launch` and `lablink register` → `lablink client register` are hard-renamed (no Typer alias for the old form). Any external script wrapping `lablink launch-client` will get \"no such command.\" The docs are updated (`docs/reference/cli.md:108`) but I don't see a CHANGELOG entry in this diff. Worth either:

- Adding short alias `@app.command(\"launch-client\", hidden=True)` and `@app.command(\"register\", hidden=True)` that delegate, **OR**
- Adding a CHANGELOG entry explicitly calling out the rename as a breaking change.

Given the CLI's age I lean toward just the CHANGELOG entry, but flagging in case external automation exists.

#### 2. (Low) `_load_config_safe` in `doctor.py:24` swallows all exceptions

```python
except Exception:
    return None
```

A typo'd `provider:` field will fall through to AWS prereqs silently. Consider narrowing to `(FileNotFoundError, yaml.YAMLError, OmegaConfBaseException)` or at least printing a `[yellow]Could not load config: ...[/yellow]` warning so the operator sees \"running AWS checks because config didn't load.\" Not blocking.

#### 3. (Low) `_resolve_manual_admin_credentials` uses literal `\"MISSING\"` sentinel

`status.py:582` treats the string `\"MISSING\"` as \"not set.\" That's brittle (what if someone legitimately picked the password `MISSING`?). If `\"MISSING\"` is the convention elsewhere in the CLI, fine — but if it's only here, prefer empty-string or `None`. Not blocking.

#### 4. (Low / clarification) `_detect_lan_ip` is really \"default-route source IP\"

`deploy_compose.py:_detect_lan_ip` connects to `8.8.8.8:80` and reads `getsockname()`. On a Tailscale / VPN host that owns the default route, this returns the 100.x.x.x address rather than a true LAN address — and BYO clients on the literal LAN can't reach it. The operator can always pass the right address manually, but a one-line comment acknowledging \"this is the source IP for the default route, which is usually but not always the LAN address\" would save someone confusion.

#### 5. (Low) `VALID_PROVIDERS` hard-codes the registry contents

`validate_config.py:28` hard-codes `(\"aws\", \"manual\")` with a comment that says \"must stay in sync with the providers registry.\" If the registry exposes `PROVIDERS.keys()` (or similar), deriving from there would avoid the rot risk. Not blocking — the comment makes the coupling explicit.

#### 6. (Nit) `start.sh` manual-mode seed depends on the venv being live for `_vncauth_blob` import

The heredoc imports `lablink_client_service.agent.kasmvnc` after the venv activate at line 8. If the install is broken, the `passwd` file won't be written and Xvnc startup fails later with a non-obvious error. The same import is required for the agent anyway, so this is fine in practice — just noting that it pushes the failure mode \"package not installed\" from a clean import error to \"Xvnc said -PasswordFile is unreadable.\"

### Test coverage

Strong. The new tests are dense and explain *why* they exist (regression guards with multi-line rationale comments rather than bare asserts). Particularly useful: the explicit `stderr=subprocess.STDOUT` kwarg check, the `--network host` removal check, the `\"--pull always\"` check, the 8-char VncAuth truncation invariant, and the `desktop_lan_direct_urlencodes_credential` test that locks `&password=a%2Fb%2Bc%3Dd` for a credential with reserved bytes.

### Security review

- VncAuth credential: 48 bits of entropy, per-session rotation, KasmVNC `brute_force_protection: 5` caps online guessing.
- URL-string credential exposure: explicitly documented (DevTools visibility), mitigated by `access_log off` + `location.replace` + per-session rotation. Acceptable for the closed-LAN deployment model this targets.
- Bearer-auth on DELETE; admin Basic + Bearer API_TOKEN on list; mode 0600 on `.env`.
- HTTP-only manual deploys (preflight rejects all TLS providers including self-signed); operators front with their own reverse proxy for public TLS.

### Recommendation

Approve once the breaking-CLI-rename gets either a hidden alias or a CHANGELOG line. Everything else is optional polish.

PR #382, opened 2026-07-17
feat(allocator): admin VNC access (peek/connect/release)

### LINKED ISSUE #378: Access VM admin button
It'd be great to have a single click button in the admin dashboard in the VM list to jump into VNC for that instance without having to enroll in the database with an email and make that instance unavailable to participants.

Use cases:
- Peek at how participants are doing (view only)
- Troubleshooting (e.g., install issues)

7174Andy:
## Summary
- Lets an admin **peek** (view-only) at a VM currently assigned to a participant, or **connect** (full control) to an idle VM for troubleshooting — neither path touches student enrollment (`useremail`)
- Reuses the existing session/cookie/proxy machinery (`prepare_browser_session`, the signed `lablink_session` cookie, `/desktop`, `/internal/proxy_auth`) rather than building a parallel system, since those routes already key off `sessionid`/`status`, never `useremail`
- New nullable `AdminReservedAt` column excludes a VM from the assignable pool while an admin is troubleshooting it; a background `AdminSessionExpiryService` force-releases forgotten reservations after a configurable timeout (default 30 min), on top of a manual Release button (in the dashboard row and on the viewer page itself)
- Scope: AWS (allocator-proxied) connectivity only; no audit logging; view-only is enforced client-side (noVNC's own flag), consistent with this admin panel's existing single-shared-credential trust model
- Closes #378

## Test plan
- [x] Full plan-touched test suite: 153 passed, 14 skipped (real-Postgres tests skip cleanly without a local Postgres in CI-equivalent conditions), 0 failures
- [x] `ruff check src tests` clean
- [x] AWS regression test (`test_desktop_aws_byte_identical`) unchanged and passing — student flow stays byte-identical
- [x] Final whole-branch review caught and this PR fixes a critical integration bug in `/internal/proxy_auth` (the nginx auth gate wasn't stripping the new cookie suffix, which would have broken both admin flows end-to-end) — covered by a new regression test
- [x] Manual smoke test against a real deployment (AWS-proxied connectivity) — recommended before merge, not yet performed

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -229,6 +229,7 @@ General application settings.
 | `admin_user` | string | `admin` | Admin username for web UI |
 | `admin_password` | string | `admin_password` | Admin password (override with `PLACEHOLDER_ADMIN_PASSWORD` or GitHub secret) |
 | `region` | string | `us-west-2` | AWS region for deployments |
+| `admin_session_timeout_minutes` | integer | `30` | Fixed duration cap for an admin's VNC troubleshooting session on an unassigned VM before it's force-released |
 
 !!! danger "Configure Passwords"
     Configure `ADMIN_PASSWORD` secret for GitHub Actions deployments, or manually replace the placeholder. See [Security](security.md#change-default-passwords).
--- a/openspec/specs/api/spec.md
+++ b/openspec/specs/api/spec.md
@@ -181,3 +181,45 @@ render the same aggregates as the admin page without recomputing.
 - **GIVEN** invalid or missing credentials
 - **WHEN** a user submits `GET /api/session-metrics/summary`
 - **THEN** HTTP 401 Unauthorized is returned
+
+### Requirement: Admin VNC Peek Endpoint
+The allocator SHALL provide an authenticated endpoint for an admin to view
+(read-only) a VM currently assigned to a participant, without modifying the
+VM's assignment.
+
+#### Scenario: Peek at an in-use VM
+- **GIVEN** valid admin credentials and a VM with an active participant session
+- **WHEN** an admin accesses `GET /admin/instances/<hostname>/peek`
+- **THEN** the admin's browser is redirected to `/desktop` in view-only mode,
+  viewing the same live session — the VM's assignment is untouched
+
+#### Scenario: Nothing to peek at
+- **GIVEN** valid admin credentials and a VM with no active participant session
+- **WHEN** an admin accesses `GET /admin/instances/<hostname>/peek`
+- **THEN** the admin is redirected to `/admin/instances` with an error indicator
+
+### Requirement: Admin VNC Connect Endpoint
+The allocator SHALL provide an authenticated endpoint for an admin to connect
+(full control) to a VM not currently assigned to any participant, for
+troubleshooting, without marking the VM assigned.
+
+#### Scenario: Connect to an idle VM
+- **GIVEN** valid admin credentials and a VM with no participant assigned
+- **WHEN** an admin submits `POST /admin/instances/<hostname>/connect`
+- **THEN** the VM is reserved (excluded from the assignable pool) and the
+  admin's browser is redirected to `/desktop` with full control
+
+#### Scenario: Race with a concurrent claim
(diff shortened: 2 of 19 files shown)


7174Andy (commented):
## Code Review

### Summary
Comprehensive review of the full PR (9 commits, base `a29e3e30`→head `16c7b60f`, +996/-36 across 19 files). Dispatched a fresh independent pass (not just my own recollection as the author) covering the entire diff, weighted toward the 5 commits added after the last whole-branch review (2 CI fixes + 3 UI-only polish commits) that hadn't been independently scrutinized yet. It verified the design doc's scope claims, traced the cookie-payload contract across all consumers, ran the affected test suites, and spot-checked the template's conditional logic across all VM states.

### Strengths
- Reusing `prepare_browser_session`, the signed `lablink_session` cookie, `/desktop`, and `/internal/proxy_auth` (all already keyed off `sessionid`/`status`, never `useremail`) kept this addition small instead of building a parallel session subsystem.
- The previously-caught Critical bug (`/internal/proxy_auth` not stripping the new cookie suffix before a UUID-column lookup) is genuinely fixed and regression-locked — confirmed there are exactly two cookie-payload consumers in the codebase and both now strip the suffix consistently.
- `admin_reserve_vm`'s atomic `UPDATE ... RETURNING` claim is proven race-safe by a real 2-thread concurrency test, not just asserted.
- XSS is handled correctly in the one place it matters: the admin-session wrapper page escapes `hostname` for both the HTML text and the form-action URL, with a dedicated regression test.
- The `instances.html` Access-column conditional (`if/elif/elif/else`) is mutually exclusive and total after three separate UI-polish commits — verified all four VM states plus the awkward combinations render correctly, and removing the two duration columns didn't break anything relying on column count/position.

### Issues

#### Critical (Must Fix)
None.

#### Important (Should Fix)
None.

#### Minor (Nice to Have)
1. **`main.py` (peek/connect/`request_vm`)** — the cookie-minting sequence (`getconn` → `get_or_create_cookie_secret` → `sign` → `set_cookie` with the same httponly/samesite/secure flags) is now copy-pasted across three routes. Worth extracting into a small helper so the cookie-hardening flags can't drift apart across call sites later.
2. **`admin_connect_vm` (`main.py`)** lacks the broad `except Exception` that `submit_vm_details`/`request_vm` has. If anything other than `RotationFailed` raises after `admin_reserve_vm` already flips `AdminReservedAt`, the admin gets a raw 500 and the VM stays reserved. Self-recoverable (dashboard shows "Admin session active" + Release, plus the 30-min sweep backstop) — hence Minor, not Important — but mirroring the existing broad-guard pattern would close the gap.
3. **`desktop.py`'s `_render_admin_session_page`** escapes `hostname` (correctly) but interpolates `ws_url` into the iframe `src` unescaped. Not exploitable today (`ws_url` is always a server-generated `proxy/{token}`), but inconsistent with the hostname handling right next to it.
4. **`instances.html`** still has the dead `<i class="bi bi-file-text"></i>` icon on the View Logs link — this page never loads an icon font, so it's always rendered nothing, and the real button styling added in a later commit didn't clean it up.
5. **`admin_session_expiry.py`**: `check_interval_seconds: int` but tests construct instances with float values (e.g. `0.05`) for fast polling — harmless at runtime (`Event.wait` accepts float), but the type hint doesn't match actual usage.

### Recommendations
- Worth a one-line manual smoke check at deploy time that nginx (in the separate `lablink-template` repo) doesn't send `X-Frame-Options`/CSP `frame-ancestors` headers that would block the same-origin iframe in the new admin-session wrapper page — the allocator Flask app itself sets no framing headers, so this is purely an infra-config question outside this repo.
- The deliberate scope boundaries from the design doc (AWS-only connectivity, no audit log, fixed 30-min cap not activity-based, client-side-only view-only enforcement, unreconciled pre-existing openspec drift) are all respected and documented — no action needed on any of those.

### Assessment
All 5 Minor findings verified against the actual source before posting. None are blocking — they're reasonable follow-up cleanup, either in this PR or a fast follow-up.

**Ready to merge: Yes**

PR #398, opened 2026-07-24
feat(allocator+cli): real-time progress tracking for launch/destroy operations

7174Andy:
## Summary

Adds real `N/total` resource-completion progress to launch/destroy operations, replacing the indeterminate spinner/elapsed-time-only treatment in both the web dashboard's operations panel and the CLI's `lablink deploy`/`lablink cleanup` commands — by parsing Terraform's streamed output instead of discarding it until the job completes.

**Rebased onto `main` now that #397 has merged.** This branch was previously stacked on `feat/admin-instances-dashboard-377` because the frontend piece (Task 8) builds on that PR's operations-history panel and job-status banner. #397's 13 commits have been dropped from this branch (they arrived on `main` as squash-merge 396935a8) and the 18 progress-tracking commits replayed on top — so the diff below is now this PR's work only.

## ⚠️ This PR changes two things #397 just shipped

Reviewers of #397 should be aware that commit `151f1806` ("match instances view-toggle layout to mockup, drop container-startup column") adjusts the `/admin/instances` page that #397 landed hours ago. Both changes came from a mockup review on this branch, and neither is required by the progress-tracking work — they're separable if you'd rather they didn't ride along:

- **`Container Startup Duration` removed from the Table view** (`<th>` + `<td>`), taking the table from 10 columns to 9. **Card view still displays it** (`Container Startup: N.Ns`), so the metric is not gone from the UI — only from the wide table, which was the column-count driver behind the horizontal-scroll fix.
- **`.view-toggle` changed from `justify-content: center` to `space-between`**, grouping the view toggle and refresh controls to opposite ends instead of centering them.

The corresponding `test_pages.py` assertions were updated in the same commit (the horizontal-scroll test's docstring no longer claims 10 columns).

### Allocator backend
- `operations` table gains `resources_total`/`resources_completed` columns (no migration — fresh-per-deployment DB), plus `OperationsDatabase.update_operation_progress` and a `finish_operation` update that snaps `resources_completed` to `resources_total` on success.
- `OperationsWorker` builds and threads a progress callback into the operation-running closure.
- `ComputeProvider` protocol (and `ManualProvider`) gain an optional `progress_callback` parameter; `ManualProvider` never calls it (it never runs Terraform).
- New `_run_streamed` helper in `AWSProvider`: streams `terraform apply`/`terraform apply <destroy-plan>` via `Popen` instead of blocking `subprocess.run`, invoking a callback as each resource's completion line streams past — with proper cleanup (kill + join) if the callback itself raises.
- `provision_hosts` computes `resources_total` from the plan JSON it already parses for the security-group audit (no new Terraform call). `destroy_hosts` gains a new `terraform plan -destroy` step to get the same exact-count treatment for destroys (small added latency, accepted trade-off for an exact count over an approximation).
- `main.py`'s `_run_launch`/`_run_destroy` forward the callback through, with a backward-compatible `None` default.

### Web dashboard
- `/admin/instances`'s operations-history panel and job-status banner render a real filled progress bar + `"N/total resources"` text once a total is known, falling back to the existing indeterminate bar when it isn't.
- Progress bars disappear once an operation reaches a terminal status — only `queued`/`running` operations show one.
- Job-status banner also gains a close button, larger text, and per-state icons + labels (so states aren't distinguished by background color alone).
- Table-view column and view-toggle adjustments — see the ⚠️ section above.

### CLI
- `AllocatorAPI._submit_and_poll` (and `launch_vms`/`destroy_vms`) gain an `on_progress` callback, read off the same 2-second poll loop that already existed — no new HTTP calls.
- `lablink deploy`/`lablink cleanup` swap their indeterminate `rich` spinner for a real `rich.progress.Progress` bar.
- Graceful degradation verified both ways: an older allocator's response simply lacks the two new fields, and both the API layer and CLI commands degrade to today's indeterminate look rather than erroring.

## Testing

Re-run after the rebase onto `main`:

- [x] Allocator suite (`PYTHONPATH=src pytest --ignore=tests/terraform`) — **752 passed, 19 skipped**
- [x] CLI suite (`PYTHONPATH=src pytest`) — **593 passed, 1 deselected** (the `test_deploy_compose.py` rich-formatting failure noted earlier is no longer present)
- [x] `ruff check src tests` clean in both packages
- [x] CI green on the rebased branch — all 3 lint jobs, all 3 test jobs, and the allocator image build pass
- [x] Manual end-to-end (launch/destroy a real VM, watch the bar move 0→N in both the web dashboard and the CLI)

Note: `tests/terraform` fails locally for want of a Terraform binary/credentials — identical failures on a clean `main` checkout, and it passes in CI.

## Related Issues

Builds on #397 (merged).

### DIFF
--- a/packages/allocator/src/lablink_allocator_service/generate_init_sql.py
+++ b/packages/allocator/src/lablink_allocator_service/generate_init_sql.py
@@ -66,7 +66,9 @@ CREATE TABLE IF NOT EXISTS operations (
     started_at TIMESTAMP,
     finished_at TIMESTAMP,
     output TEXT,
-    error TEXT
+    error TEXT,
+    resources_total INTEGER,
+    resources_completed INTEGER
 );
 
 CREATE INDEX idx_operations_created_at ON operations(created_at);
--- a/packages/allocator/src/lablink_allocator_service/main.py
+++ b/packages/allocator/src/lablink_allocator_service/main.py
@@ -10,6 +10,7 @@ from datetime import datetime
 import re
 import atexit
 from functools import wraps
+from typing import Callable, Optional
 
 from flask import (
     Flask,
@@ -660,13 +661,17 @@ def launch():
         "deployment_name": getattr(cfg, "deployment_name", "lablink"),
     }
 
-    def _run_launch() -> str:
+    def _run_launch(
+        progress_callback: Optional[Callable[[int, int], None]] = None,
+    ) -> str:
         """Runs on OperationsWorker's background thread, not the request
         thread. Reformats known failure types into RuntimeError with the
         same user-facing text the old synchronous route used, so that
         text ends up in the operation's `error` column."""
         try:
-            result = provider.provision_hosts(count=total_vms, spec=spec)
+            result = provider.provision_hosts(
+                count=total_vms, spec=spec, progress_callback=progress_callback,
+            )
         except SGAuditFailure as exc:
             raise RuntimeError(
                 f"Security-group audit refused the plan: {exc}"
@@ -722,7 +727,9 @@ def destroy():
             return jsonify({"status": "error", "error": error_msg}), 405
         return redirect("/admin/instances?error=destroy_unsupported")
 
-    def _run_destroy() -> str:
+    def _run_destroy(
+        progress_callback: Optional[Callable[[int, int], None]] = None,
+    ) -> str:
         """Runs on OperationsWorker's background thread, not the request
         thread."""
         # Seal any open session-metrics rows before tearing down VMs, so the
@@ -738,7 +745,9 @@ def destroy():
         # destroy_hosts ignores the handles arg (terraform destroy operates
         # on the whole workspace); skip the list_hosts() call.
         try:
-            result = provider.destroy_hosts([])
+            result = provider.destroy_hosts(
+                [], progress_callback=progress_callback,
+            )
         except FileNotFoundError as e:
             # No terraform.runtime.tfvars → no client VMs were ever launched.
             raise RuntimeError(str(e)) from e
--- a/packages/allocator/src/lablink_allocator_service/operations.py
+++ b/packages/allocator/src/lablink_allocator_service/operations.py
@@ -42,7 +42,7 @@ class OperationsWorker:
     def submit(
         self,
         op_type: str,
-        fn: Callable[[], str],
+        fn: Callable[..., str],
         params: Optional[str] = None,
         created_by: Optional[str] = None,
     ) -> int:
@@ -63,10 +63,15 @@ class OperationsWorker:
         ).start()
         return operation_id
 
-    def _run(self, operation_id: int, fn: Callable[[], str]) -> None:
+    def _run(self, operation_id: int, fn: Callable[..., str]) -> None:
+        def _progress_callback(completed: int, total: int) -> None:
+            self.database.update_operation_progress(
+                operation_id, completed, total
+            )
+
         try:
             self.database.start_operation(operation_id)
-            output = fn()
+            output = fn(_progress_callback)
         except Exception as e:
             logger.error(
                 "Operation #%d failed: %s", operation_id, e, exc_info=True
--- a/packages/allocator/src/lablink_allocator_service/operations.py
+++ b/packages/allocator/src/lablink_allocator_service/operations.py
@@ -64,10 +64,28 @@ class OperationsWorker:
         return operation_id
 
     def _run(self, operation_id: int, fn: Callable[..., str]) -> None:
+        progress_disabled = False
+
         def _progress_callback(completed: int, total: int) -> None:
-            self.database.update_operation_progress(
-                operation_id, completed, total
-            )
+            # Best-effort: a transient DB failure here must never abort a
+            # live terraform apply/destroy (which _run_streamed would do by
+            # killing the subprocess if this raised). Once a write fails,
+            # stop trying for the rest of this operation rather than
+            # hammering an already-struggling database on every resource.
+            nonlocal progress_disabled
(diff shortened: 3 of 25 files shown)


7174Andy (commented):
## Code Review

### Summary
Reviewed the full diff (17 commits, 25 files) plus the latest commit (`8c83e0b5`), which addresses 3 correctness findings from an independent (codex) pre-merge review. Ran both full test suites, `ruff check`, and manually re-verified the key logic paths (progress-callback error handling, streamed-subprocess cleanup, resources_total computation, CLI progress-field guards). This is a well-scoped, carefully-tested feature; I have no blocking issues left.

### What the latest commit fixed
- **`operations.py`**: a DB write failure while recording progress (e.g. transient pool exhaustion) no longer propagates into `_run_streamed`, which would otherwise `kill()` a live `terraform apply`/`destroy` over an unrelated, transient failure. The callback now catches, logs, and disables further progress writes for that operation — verified with a new test (`test_progress_callback_write_failure_does_not_abort_operation`) that a raising `update_operation_progress` still lets `fn` complete and the operation finish `succeeded`.
- **`launch.py` / `deploy.py`**: `_on_progress` previously gated only on `total` truthiness; a response with `resources_total` present but `resources_completed=None` (or vice versa) would have rendered the literal string `"None/5 resources"` in the CLI progress description. Now both fields are required to be non-`None` before updating. New mixed-field tests added to both `test_launch.py` and `test_deploy.py`.
- **`test_operations.py`**: `test_submit_does_not_block_caller_on_slow_fn`'s `slow_fn()` took zero args, but `_run` always calls `fn(progress_callback)` — the test was passing because `fn` crashed immediately with `TypeError`, not because it ran asynchronously. Fixed the signature so the timing assertion is meaningful again.

All three were verified against the actual code before fixing (not taken at face value) — file:line evidence is in the local session log if useful for the record.

### Strengths
- `_run_streamed` (`providers/aws.py`) correctly mirrors `subprocess.run`'s cleanup-on-exception contract: `except BaseException: proc.kill(); raise` / `finally:` closes stdout, joins the stderr-drain thread, and calls `proc.wait()` — verified via `test_run_streamed_kills_process_and_still_cleans_up_when_callback_raises`, which spies on the real thread object to assert it's actually joined, not just fire-and-forget.
- The duration-regex bug (`_CREATE_COMPLETE_RE`/`_DESTROY_COMPLETE_RE` only matching `\d+s`, silently missing Terraform's `5m2s`/`1h2m3s` formats) was caught via real E2E testing and fixed with regression tests built from the user's actual log output (5 VMs, 5–7 min destroy each) — a good example of a bug that unit tests alone wouldn't have caught.
- Graceful degradation is threaded consistently through every layer: `main.py`'s `_run_launch`/`_run_destroy` default `progress_callback=None`, the CLI's `on_progress` callbacks handle `(None, None)` explicitly, and the frontend's `renderOperationsHistory`/`renderOperationBanner` gate the whole progress block on `op.resources_total` being truthy (not just interpolating blindly) — so an allocator predating this feature, or a partial response, degrades to the old indeterminate-spinner look everywhere rather than crashing or rendering "None".
- XSS handling in `instances.html` is applied consistently with the file's existing convention: free-text fields (`op.error`, `op.output`, `op.created_by`) go through `escapeHtml()`; server-controlled enum/int fields are interpolated raw.
- Test quality is high throughout — e.g. `test_operations_db.py` uses full `"".join(query.split())` equality against a shared `_FINISH_OPERATION_QUERY` constant rather than brittle substring checks, matching the file's established pattern.

### Issues Found

#### Minor
- **`providers/aws.py` `resources_total` computation** (`provision_hosts`/`destroy_hosts`): both count via `"create" in actions` / `"delete" in actions` against `plan_json["resource_changes"]`, which correctly excludes `no-op`/`update`/`read` entries — but neither `test_aws_provision_hosts.py::test_provision_hosts_reports_progress_via_callback` nor the destroy equivalent exercises a plan with a mixed action list (only pure all-`create`/all-`delete` plans). This logic is already exercised for real in production (a launch's plan always includes `no-op` entries for pre-existing resources like the security group), and the live destroy the user tested landed on an exact 35-resource count with no overshoot — so I'm confident the logic is correct, just not explicitly unit-tested for the mixed case. Not blocking; a follow-up test would be a nice-to-have.
- **Provider protocol change is a soft break for third-party `lablink.providers` plugins** (`providers/protocol.py`, `main.py`): `provision_hosts`/`destroy_hosts` now always receive `progress_callback=` as a keyword from `main.py`, regardless of whether the callee's provider class declares that parameter. Only `aws`/`manual` are registered today (both updated in this PR), so nothing is broken in this repo right now — but any hypothetical external provider implementing the old 2-arg signature would get a `TypeError`. Worth a CHANGELOG note for plugin authors; not worth an adapter shim given zero known third-party providers exist today.
- The PR description's own migration note ("no migration — fresh-per-deployment DB") is worth double-checking against how the allocator is actually upgraded in the field (vs. a from-scratch deploy) — I traced `start.sh` and found no persistent Postgres volume mount, which supports the "fresh per container" claim, but I'd flag this as a place to stay alert if an in-place upgrade path is ever introduced later.

### Testing
- Allocator: `752 passed, 19 skipped` (excluding `tests/terraform`, which needs live AWS credentials/region env vars not present in this environment — confirmed pre-existing via `git stash`, unrelated to this PR).
- CLI: `592 passed`, 1 pre-existing failure in `test_deploy_compose.py` (rich/ANSI terminal-width rendering mismatch, confirmed pre-existing via `git stash`, unrelated to this PR).
- `ruff check` clean on all files touched by the latest commit.

### Recommendation
Approve pending #397 merging first (this branch is stacked on it, per the PR description). No blocking issues.

PR #400, opened 2026-07-27
fix(cli): report auth failures instead of empty results

7174Andy:
## Summary

- `lablink status` reported **"No client VMs found"** plus a full cost table when the actual problem was an unauthenticated caller — a silent wrong answer, not an error.
- The same bug class made a rejected admin login surface as a bare **"HTTP 401"** in `status` (manual provider), `stats`, and `export-metrics`.
- Root cause in both: helpers collapsing *"the query failed"* into *"nothing exists"*.

This was found from a real report, and the reporter's machine turned out to have genuinely invalid AWS credentials (`InvalidClientTokenId`) — `aws sts get-caller-identity` said so plainly while `lablink status` hid it entirely.

### AWS credential failures (`utils.py`, `status.py`, `logs.py`)

- `AwsQueryError` carrying `is_auth`, so credential/permission problems get remediation steps while throttling and endpoint errors are reported verbatim.
- `_classify_aws_error()` maps `NoCredentialsError`, `PartialCredentialsError`, SSO token errors (`TokenRetrievalError`, `UnauthorizedSSOTokenError`, `SSOTokenLoadError`), `ProfileNotFound`, and the auth-flavored `ClientError` codes (`AuthFailure`, `UnauthorizedOperation`, `AccessDenied*`, `ExpiredToken*`, `RequestExpired`, `InvalidClientTokenId`, `SignatureDoesNotMatch`, `UnrecognizedClientException`).
- `aws_credentials_error()` — a **quiet** STS probe. Deliberately not reusing `setup.check_credentials`, which prints a red wall and `SystemExit`s; that's why `doctor._check_aws_credentials` has to catch `SystemExit`.
- **`query_ec2_instances` now raises** instead of returning `[]`. An empty list means the query succeeded and matched nothing.
- `status` probes credentials up front and prints one banner (reason, remediation, region/profile) instead of three misleading empty sections. DNS/HTTP/SSL checks still run — they need no AWS. Costs render marked as fallback and skip Pricing API calls that can only fail.
- `_render_client_vms` also catches `AwsQueryError` directly, which covers valid-credentials-but-missing-`ec2:DescribeInstances` — something the upfront probe cannot see.
- `logs` catches `AwsQueryError` and exits 1 instead of tracebacking (required: `list_all_vms` can now raise).

### Rejected admin logins (`status.py`, `stats.py`, `export_metrics.py`)

Each now names the rejected admin user and where its credentials come from, via a shared `print_admin_credentials_hint()`. Non-401 codes keep their previous generic wording, since credential guidance would be actively misleading there.

`stats.py` also needed its handlers reordered: it caught `(HTTPError, URLError)` in one clause, and since `HTTPError` subclasses `URLError`, a 401 branch added there would have been unreachable.

## Testing

- **630 CLI tests pass** (1 integration deselected), `ruff check` clean across all three packages. Every new line is covered; remaining gaps in the touched files are all pre-existing.
- Written test-first — each new test was confirmed failing against the old code (e.g. `assert 'admin' in 'HTTP 401: Unauthorized\n'`).
- Updated the two existing `test_*_returns_empty` cases in `test_utils.py`, which encoded the old swallow-and-return-`[]` contract.
- Manual verification of the AWS path: bogus credentials, genuinely absent credentials, an exported-but-empty `AWS_PROFILE`, and `lablink logs` (exit 1, no traceback).
- Confirmed against the allocator that `/api/session-metrics/summary`, `/api/export-metrics`, and `/api/v1/clients` all answer **401** for *both* missing and wrong-password Basic auth (two were already pinned by existing allocator tests; the third I verified with a throwaway test).
- Confirmed via a real local socket that `urllib` surfaces flask_httpauth's 401 — including its `WWW-Authenticate: Basic realm=...` header — as `HTTPError.code == 401`, and that the `Authorization` header **survives a 302**. That last check matters: had it been dropped, a Funnel-fronted `http://` allocator would 401 with *correct* credentials and this message would send people to edit config files that were fine.

## Design Decisions

- **Raise rather than return a tuple.** Only two production call sites exist (`status.py`, `logs.py`), so raising keeps the honest signal without back-compat wrappers whose whole job is to re-hide the error.
- **One upfront probe rather than three per-section errors.** Terraform state, VM inventory, and pricing all degrade to empty on a credential failure; reporting once at the top and labelling the affected sections reads better than the same error three times. Costs one extra STS call per `status` run.
- **Costs stay visible under dead credentials,** marked as fallback, rather than being hidden — the static estimate is still useful.
- **Hints are pre-split into short lines.** Rich wraps at console width without carrying the indent onto continuation lines; the first attempt rendered raggedly, which is why the output was eyeballed rather than assumed.
- **`is_auth` gates remediation** so `aws configure` is never suggested over a throttling error.

## Not included

- `get_terraform_outputs` still discards Terraform's stderr — parsing that is a separate job; the new banner already explains the empty state.
- `doctor._check_aws_credentials` could now drop its `except SystemExit` hack and call `aws_credentials_error`; left alone to keep this diff scoped.
- `logs.py:64` already had a 401 branch. It's terser than the new wording, but it returns an error dict into the TUI rather than printing, so the multi-line hint would need that panel reworked.

---

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -139,6 +139,12 @@ Shows four sections:
 3. **Client VMs** — per-VM state and current hourly burn rate.
 4. **Cost Estimate** — daily and monthly dollar estimates, pulled from the AWS Pricing API with a fallback table.
 
+If your AWS credentials are missing, expired, or lack permission, an **AWS
+credentials** section is printed first with the reason and how to
+authenticate. The Terraform state and Client VM sections then say they're
+unavailable rather than showing an empty result, and costs fall back to the
+built-in price table.
+
 | Option | Description |
 |---|---|
 | `-c`, `--config PATH` | Path to `config.yaml`. |
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -139,11 +139,15 @@ Shows four sections:
 3. **Client VMs** — per-VM state and current hourly burn rate.
 4. **Cost Estimate** — daily and monthly dollar estimates, pulled from the AWS Pricing API with a fallback table.
 
-If your AWS credentials are missing, expired, or lack permission, an **AWS
-credentials** section is printed first with the reason and how to
-authenticate. The Terraform state and Client VM sections then say they're
-unavailable rather than showing an empty result, and costs fall back to the
-built-in price table.
+If your AWS credentials are missing or expired, an **AWS credentials**
+section is printed first with the reason and how to authenticate. The
+Terraform state and Client VM sections then say they're unavailable rather
+than showing an empty result, and costs fall back to the built-in price
+table.
+
+If the credentials are valid but lack a required IAM permission — or any
+other AWS error occurs — the affected section reports that error in place,
+with guidance to fix the policy rather than to re-authenticate.
 
 | Option | Description |
 |---|---|
--- a/packages/cli/src/lablink_cli/commands/export_metrics.py
+++ b/packages/cli/src/lablink_cli/commands/export_metrics.py
@@ -23,6 +23,7 @@ from rich.console import Console
 from lablink_cli.api import authenticated_json_request
 from lablink_cli.commands.utils import (
     get_allocator_url,
+    print_admin_credentials_hint,
     resolve_admin_credentials,
 )
 from lablink_cli.deployment_metrics import load_all_metrics
@@ -65,7 +66,14 @@ def _export_client_metrics(
             url, admin_user, admin_pw, ssl_provider=cfg.ssl.provider
         )
     except HTTPError as e:
-        console.print(f"[red]HTTP {e.code}: {e.reason}[/red]")
+        if e.code == 401:
+            console.print(
+                f"[red]The allocator rejected admin user "
+                f"'{admin_user}' (HTTP 401).[/red]"
+            )
+            print_admin_credentials_hint()
+        else:
+            console.print(f"[red]HTTP {e.code}: {e.reason}[/red]")
         raise SystemExit(1) from e
     except URLError as e:
         console.print(f"[red]Connection error: {e.reason}[/red]")
--- a/packages/cli/src/lablink_cli/commands/logs.py
+++ b/packages/cli/src/lablink_cli/commands/logs.py
@@ -14,10 +14,12 @@ from lablink_allocator_service.conf.structured_config import Config
 
 from lablink_cli.api import authenticated_json_request
 from lablink_cli.commands.utils import (
+    AwsQueryError,
     get_allocator_url,
     get_deploy_dir,
     get_terraform_outputs,
     list_all_vms,
+    print_aws_error,
     resolve_admin_credentials,
 )
 
@@ -402,7 +404,12 @@ def run_logs(cfg: Config) -> None:
         f"'{cfg.deployment_name}' ({cfg.environment})...[/dim]"
     )
 
-    vms = list_all_vms(cfg)
+    try:
+        vms = list_all_vms(cfg)
+    except AwsQueryError as e:
+        print_aws_error(e, prefix="Could not list VMs")
+        raise SystemExit(1) from e
+
     if not vms:
         console.print(
             f"[red]No running VMs found for deployment "
--- a/packages/cli/src/lablink_cli/commands/stats.py
+++ b/packages/cli/src/lablink_cli/commands/stats.py
@@ -16,6 +16,7 @@ from rich.table import Table
 from lablink_cli.api import authenticated_json_request
 from lablink_cli.commands.utils import (
     get_allocator_url,
+    print_admin_credentials_hint,
     resolve_admin_credentials,
 )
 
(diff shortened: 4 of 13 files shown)


7174Andy (commented):
## Code Review

**Caveat up front:** this is a self-review by the author, so it is weaker than a second pair of eyes — I am checking my own assumptions. It should not substitute for human review before merge.

All 7 CI checks pass (lint + test for all three packages, docs). 630 CLI tests, no coverage gate on the CLI job.

### Issues Found

#### Important

**1. `status.py:686` points operators at a file that does not exist.**

The new 401 message says to check `app.admin_user` / `app.admin_password` *"in your lablink.yaml"*. The canonical config is `~/.lablink/config.yaml` (`app.py:19`, `DEFAULT_CONFIG`). Grepping the repo, `lablink.yaml` appears in exactly three places: a comment in `wizard.py:1447`, a pre-existing message at `stats.py:71`, and this new line. It is referenced nowhere in `docs/`.

This is the worst possible place for a wrong filename — the entire purpose of the message is to point someone at the right file, and instead it sends them hunting for one that was never there. Should read `~/.lablink/config.yaml` (or "your config", as the `stats`/`export-metrics` hint already does). Note `stats.py:71` has the same pre-existing error and could be fixed in passing.

**2. `run_status` gates too broadly on *any* probe failure, not just auth failures.**

```python
aws_down = aws_error is not None
_render_client_vms(cfg, aws_unavailable=aws_down)
```

If the STS probe fails for a non-auth reason — a transient endpoint blip, or STS blocked by a proxy/VPC endpoint policy while EC2 itself is reachable — this now **skips the EC2 inventory entirely**, where the old code would at least have attempted it. That is a small capability regression in the one case where the probe is wrong about the world.

The precise gate is `aws_error.is_auth`: let non-auth probe failures fall through to the real queries, which now report their own errors properly via the `except AwsQueryError` branch in `_render_client_vms`. That branch already exists for exactly this purpose and is currently unreachable whenever the probe has failed for any reason.

Relatedly, `_render_aws_credentials_error`'s closing line — *"Terraform state, VM inventory and live pricing are unavailable until this is fixed"* — overstates for a non-auth failure.

#### Moderate

**3. The remediation text is wrong for permission failures, which this PR advertises as covered.**

`AccessDenied`, `AccessDeniedException`, and `UnauthorizedOperation` classify as `is_auth=True`, which prints:

```
aws configure
export AWS_ACCESS_KEY_ID=... AWS_SECRET_ACCESS_KEY=...
aws sso login
```

None of those fix a valid identity that is missing `ec2:DescribeInstances` — that needs an IAM policy change. The PR description specifically cites permission failures as a benefit of catching `AwsQueryError` in `_render_client_vms`, so the guidance is misleading precisely where value was claimed. Either split the flag (`is_auth` vs `is_permission`) with distinct remedies, or add an IAM line to the remedy list.

#### Minor

- **Two `Console()` instances.** `print_aws_error` writes via `utils.console` while the surrounding report uses `status.console`. Harmless today (both to stdout, Rich flushes eagerly), but they would diverge if width or output capture were ever configured on one.
- **`get_terraform_outputs` has no `subprocess.run` timeout** and is now invoked while credentials are known-dead, where an S3 backend read will retry before failing. The missing timeout is pre-existing, but the new flow makes the slow path more reachable.
- **New tests copy an odd pre-existing patch idiom:** `patch("...commands.utils._get_session", create=True)` is a no-op decoration — the effective patch is the inner `setup._get_session`, since `query_ec2_instances` imports it at call time. Consistent with neighbours, but worth knowing it is doing nothing.

### Strengths

- Root cause fixed at the source rather than papered over at the call site; `query_ec2_instances` raising means the conflation cannot silently return.
- Tests written failing-first, and the two existing tests that encoded the old swallow-and-return-`[]` contract were updated rather than deleted.
- The 401 reachability was verified end-to-end rather than assumed: 401 confirmed for both missing and wrong-password Basic auth on all three endpoints, and the `Authorization` header confirmed to survive Funnel's 302 — without which this message would fire on *correct* credentials.

### Recommendation

Fix **1** before merge — it is a one-word change and actively misleads. **2** is worth fixing while the code is fresh, since it silently disables a code path added in this same PR. **3** can be a follow-up if scope is a concern.

### Verification gaps worth naming

The valid-credentials happy path was exercised only by unit tests; the ambient credentials on the authoring machine are invalid (`InvalidClientTokenId`), so no live end-to-end run against working AWS credentials was possible. The manual-provider 401 was verified against a stubbed 401, not a running allocator container.

7174Andy:
Addressed finding **3** in 685f0823 — all three review findings are now resolved.

`AwsQueryError` carries two mutually exclusive flags instead of one overloaded `is_auth`:

- **`is_auth`** — identity could not be established (absent, expired, or invalid credentials). Credential remedies apply.
- **`is_permission`** — identity is known but not allowed to make this call. IAM guidance applies, because re-authenticating as the same principal changes nothing.

Neither set still means "report verbatim, offer no advice".

`AccessDenied`, `AccessDeniedException`, and `UnauthorizedOperation` moved to the authorization set. Output for a real `UnauthorizedOperation` (valid credentials, missing `ec2:DescribeInstances`) — the case this PR's `except AwsQueryError` branch exists to catch and which the STS probe cannot see:

```
Client VMs
  Could not query EC2: AWS denied the request: UnauthorizedOperation — You are
not authorized to perform this operation. User:
arn:aws:iam::711387140753:user/ci is not authorized to perform:
ec2:DescribeInstances
  These credentials are valid but lack permission for this call.
  Grant the calling identity the action named above (for example
  ec2:DescribeInstances), or switch to a role or profile that has it.
```

AWS's own message already names the missing action, so the guidance points at it rather than guessing. The permission text deliberately avoids naming `aws configure` even to dismiss it — a skimming operator would try it anyway.

Two notes on scope:

- **The `sts:GetCallerIdentity` probe can, in principle, return `AccessDenied`** under an SCP or permission boundary. With the finding-2 gate now keyed on `is_auth`, that case falls through to the real queries rather than pre-emptively skipping them, which is the behaviour we want.
- **Corrected the docs paragraph I added earlier in this branch.** It claimed permission failures produce the credentials banner and that dependent sections report unavailable — both untrue after narrowing the gate in 43d5f8b5. Stale docs written in the same PR are still stale docs.

637 CLI tests pass, `ruff check` clean across all three packages. Five tests added for the split, all confirmed failing beforehand (14 failures against the previous commit, including the parametrized authorization codes).

Issue #404, opened 2026-07-28
mesh_overlay: re-registering a client mints a new Tailscale node with a -N suffix, stranding the allocator's stored overlay_hostname


7174Andy:
## Summary

Every re-registration (or container recreate) of a `mesh_overlay` client mints a **new Tailscale node**. Tailscale will not hand a MagicDNS name to a new node while an existing node — even an offline one — still holds it, so it appends a numeric suffix: `lablink-client-local-gpu-1` becomes `lablink-client-local-gpu-1-1`, then `-2`, and so on.

The allocator, however, stores the name the client *asked for* in `provider_metadata.overlay_hostname` and builds the client's address from it. After the first re-register, that stored name resolves to the **previous, now-dead node**. Every allocator → client call goes to a black hole, and the client is permanently marked `Unhealthy`, so it can never be assigned to a participant or reached over VNC.

## Impact

Confirmed live on a `provider: manual` / `connectivity: mesh_overlay` deployment (`sleap-lablink`, tailnet `tail9f6f81.ts.net`) on 2026-07-28:

- **Participant assignment dies.** `assign_vm` skips `healthy = 'Unhealthy'` rows (`packages/allocator/src/lablink_allocator_service/db/vms.py:485`), so the pool is empty.
- **Admin VNC dies.** Admin Connect can never reach the `:6080` proxy hop — it first POSTs `/api/session/start` to the client agent on `:7070` to mint the session, and that dial times out. Result is a redirect to `?vnc_error=rotation_failed`.
- **It does not self-heal.** The `Unhealthy` flag is sticky: `check_gpu.py:72` only POSTs GPU health *on change*, and the client already reported `Healthy` once before the failure, so it never re-reports. The recovery sweep logs `provider manual cannot recover hosts; leaving <host> failed`. A single transient failure permanently wedges a single-client deployment. (Related known gap.)

The client itself is completely healthy throughout — it joins the tailnet, runs its startup script, serves KasmVNC on `:6080` and the agent on `:7070`, and reports `status=running`. Nothing in the client logs indicates a problem, which makes this hard to diagnose from the client side.

## Root cause chain

1. `_build_docker_run` (`packages/cli/src/lablink_cli/commands/register.py:412`) mounts **no volume at `/var/lib/tailscale`**. tailscaled's state lives only in the container's ephemeral filesystem, so every start generates a new node key → a new machine in the tailnet.
[...]

PR #414, opened 2026-07-29
fix(allocator): stop one rotation timeout from permanently wedging a client (#404)

7174Andy:
## Summary

- A **successful** password rotation now clears an allocator-set `Unhealthy` flag, making Admin Connect the automatic recovery path.
- Adds `POST /admin/instances/<hostname>/clear-unhealthy` and a "Clear Unhealthy" button as the manual escape hatch.
- **PR 3 of 3** for #404. Stacked on #413 — review #412 and #413 first; this PR's own diff is allocator-only.

## The bug

A failed rotation marks a client `Unhealthy`, and `assign_vm` skips those rows — so a single transient connect timeout empties the assignable pool permanently.

The client cannot undo it. `check_gpu.py` only POSTs on **change**, and on a GPU-less BYO box it reports `'N/A'` once with `break_now=True` and then **exits its loop entirely**, so the allocator's write can never be overwritten. The recovery sweep logs `provider manual cannot recover hosts`. Raw SQL was the only way out.

This is what turned a recoverable blip into a dead deployment in #404.

## Changes Made

- `db/vms.py`: `clear_unhealthy()`.
- `routes/admin_sessions.py`: clear on rotation success; new `clear-unhealthy` route.
- `routes/public.py`: clear on rotation success.
- `templates/instances.html`: "Clear Unhealthy" button, rendered only on `Unhealthy` rows.

## Design Decisions

**Why clearing on rotation *success* is safe.** Admin Connect already ignores `healthy` and targets one specific host, so it was the natural retry path. A successful `prepare_browser_session` means the rotation POST to the client's agent succeeded, which proves allocator → client reachability — the exact direction whose failure set the flag. No participant can ever be handed a client that was not just proven reachable, which is what the pool predicate protects. Also wired into the participant path's rejoin branch, which matches on `status='running'` rather than going through `assign_vm` and so can land on a row still marked `Unhealthy`.

**Why `NULL`, not `'Healthy'`.** The allocator marks a client `Unhealthy` when it cannot *reach* it, which says nothing about the GPU — and `'N/A'` (no nvidia-smi, normal on a BYO box) is a legitimate client-reported value we must not overwrite with a lie. `assign_vm`'s predicate is already `(healthy IS NULL OR healthy <> 'Unhealthy')`, so NULL is assignable. The `WHERE healthy = 'Unhealthy'` guard keeps it from touching anything else. The button is labelled "Clear Unhealthy" rather than "Mark healthy" for the same reason.

**Why the button sits outside the `vm_actions` if/elif chain.** That chain is mutually exclusive, so another `elif` would have *replaced* the Connect button on exactly the rows that need it most — and a successful Connect is the automatic clear.

**Alternative rejected: clear on heartbeat.** Fully automatic, but the heartbeat is client → allocator (outbound) and does not prove allocator → client reachability, the direction that actually failed. It would resurrect a genuinely unreachable client straight into the pool.

## Testing

- 6 new tests: DB (NULL + guard), admin connect success clears, rotation failure still marks and does **not** clear, route clears + redirects, route requires auth, template shows the button only on `Unhealthy` rows while still offering Connect.
- Allocator 767 passed / 19 skipped; CLI 674; client 137; `ruff check` clean.
- **Verified against a real Postgres 15**: `clear_unhealthy` nulled the `Unhealthy` row while leaving `'Healthy'` and `'N/A'` untouched, and the cleared row then satisfied `assign_vm`'s predicate — i.e. genuinely back in the pool.
- The `admin_connect_success_clears_unhealthy` test was written after its implementation, so rather than assume it was meaningful I temporarily disabled the `clear_unhealthy` call, confirmed the test fails, then restored and re-ran. That assertion is verified to have teeth.

## Related Issues

Refs #404 — closes item 3. Items 1 and 2 are #412 and #413; item 4 (MagicDNS resolution inside the Flask process) is deliberately deferred to its own issue, since pointing the allocator container's `dns:` at `100.100.100.100` would break all DNS in that container until the sidecar's tailscaled is up.

---

### DIFF
--- a/packages/allocator/src/lablink_allocator_service/db/vms.py
+++ b/packages/allocator/src/lablink_allocator_service/db/vms.py
@@ -251,6 +251,44 @@ class VmDatabase:
             return None
         return row[0]
 
+    def set_overlay_hostname(
+        self, *, hostname: str, overlay_hostname: str
+    ) -> bool:
+        """Reconcile a mesh-overlay client's recorded overlay hostname with
+        the name Tailscale actually assigned it.
+
+        `tailscale up --hostname=X` does NOT guarantee the node is named X:
+        if an earlier (possibly offline) node still holds X, Tailscale
+        appends a numeric suffix and still exits 0. The name we recorded at
+        registration is therefore only a *request*, and after any
+        re-registration it can point at a dead node — every allocator ->
+        client dial then black-holes (lablink#404). The client reports the
+        real name after joining; this persists it.
+
+        Guarded on the key already existing so this can only ever correct a
+        mesh-overlay client: a lan_direct or AWS client cannot have an
+        `overlay_hostname` injected into its provider_metadata. Returns True
+        when a row was updated, False for an unknown or non-overlay host.
+        """
+        query = (
+            f"UPDATE {self.table_name} "
+            f"SET provider_metadata = jsonb_set("
+            f"    COALESCE(provider_metadata, '{{}}'::jsonb), "
+            f"    '{{overlay_hostname}}', to_jsonb(%s::text), true) "
+            f"WHERE hostname = %s "
+            f"  AND provider_metadata ? 'overlay_hostname';"
+        )
+        with self._cursor as cursor:
+            try:
+                cursor.execute(query, (overlay_hostname, hostname))
+                return cursor.rowcount > 0
+            except Exception as e:
+                logger.error(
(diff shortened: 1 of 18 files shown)


7174Andy:
Closed automatically by GitHub when its base branch (#413) was merged and deleted — not an intentional close. Superseded by #415, which carries the identical commit rebased onto `main`.

PR #416, opened 2026-07-29
feat(allocator): add relay connectivity for off-LAN clients without Tailscale

7174Andy:
## Summary

- Adds `RelayClientConnectivity` — a third `ManualProvider` connectivity option for BYO clients whose network **won't carry Tailscale at all** (blocked protocol, or a no-third-party-VPN policy).
- Inverts the connection direction: instead of the allocator dialing *in* to a stable address, the client dials **out** through an `frp` tunnel, and a per-client `frpc`-visitor subprocess inside the allocator's container pulls that client's KasmVNC (`:6080`) and agent (`:7070`) ports down to a dedicated loopback alias.
- **Allocator-side foundation only.** Hand-curl + hand-run-`frpc` testable on its own; client-image/CLI wiring and `deploy_compose`/wizard support are separate follow-ups.

## Changes Made

| Area | Change |
|---|---|
| `providers/protocol.py` | New `requires_frp_check` capability flag on `ClientConnectivity` (set `False` on the three existing classes) |
| `relay_manager.py` *(new)* | `start_visitor` / `stop_visitor` / `frp_status` — per-client `frpc`-visitor subprocess lifecycle plus an `frps` liveness probe |
| `db/vms.py` | `get_relay_alias`, `allocate_relay_alias_octet` (both inside the existing `provider_metadata` JSONB — no migration) |
| `providers/connectivity/relay.py` *(new)* | `RelayClientConnectivity` + `_resolve_relay_alias`; registered as `_CONNECTIVITY_BUILTIN["relay"]` |
| `conf/structured_config.py` | `relay_server_addr`, `frps_bind_port`, `frps_auth_token` |
| `validate_config.py` | Admit `"relay"`; require `relay_server_addr`; reject a weak/empty `frps_auth_token` via a new `is_weak_frps_token` |
| `routes/health.py` | `"frp"` entry in `/api/health` when `requires_frp_check` is set |
| `routes/registration.py` | Relay registration path (alias + STCP secret minting, visitor spawn) and a feature-detected `cleanup_client_identity` hook on unregister |
| `Dockerfile` / `start.sh` | `frp` binaries; `frps` startup gated on `FRPS_AUTH_TOKEN` |

## Testing

- **842 passed, 19 skipped, 0 failed** (`pytest tests/ --ignore=tests/terraform`). Baseline before this branch was 767 passed, so this adds 75 tests. `tests/terraform` is excluded locally only — it shells out to the `terraform` binary and needs AWS credentials; CI runs the full tree.
- `ruff check src tests` — clean.
- Written test-first throughout: every test was run and confirmed failing for its intended reason before the implementation landed.
- **Manual verification of the image:** `docker build` succeeds; with `FRPS_AUTH_TOKEN` set, `frps` starts and listens on `:7000` with a correctly rendered `/tmp/frps.toml`. Without it, `frps` does **not** start and no config is written — confirming the gate is a true no-op for `lan_direct`/`mesh_overlay`.
- **Hand-tested end to end** (see the section below) — this is no longer outstanding.

## Design Decisions

- **Loopback-alias addressing (the load-bearing trick).** Each relay client gets its own `127.0.0.<n>`, and its visitor binds `<alias>:6080`/`<alias>:7070`. Because `fallback_fn` returns the alias as an ordinary "private IP" string, `client_session.prepare_browser_session`'s existing `f"{ip}:6080"` building keeps working — so there are **zero changes** to `client_session.py`, `routes/desktop.py`, `internal_proxy_auth.py`, or `lablink-nginx.conf`. This needs frp's **STCP** proxy type specifically: a plain TCP proxy binds via a server-wide `proxyBindAddr` with no per-proxy override, which could not put different clients on different aliases.
- **No sidecar.** Unlike Tailscale (which needs `NET_ADMIN`/`/dev/net/tun` and was deliberately isolated), `frp` needs no elevated capabilities, so `frps` and the visitors run in the allocator's own container — the same way `start.sh` already backgrounds Flask before foregrounding nginx.
- **One subprocess per client** rather than one shared `frpc` with N hot-reloaded `[[visitors]]` blocks, so correctness doesn't depend on frp's config-reload being race-free under concurrent registrations.
- **`relay_secret_key` is stored in plaintext, deliberately.** Unlike `client_secret` (only ever verified, so hashing is right), the STCP secret is **symmetric** — the allocator's own visitor must reproduce it verbatim. Blast radius is scoped by per-proxy `serverName` uniqueness; revocation is killing the visitor, not invalidating a hash.
- **Aliases are never recycled** (`MAX(octet)+1`). A dedicated allocation table would need migration machinery this repo avoids. Note the real bound: only the final octet varies, so this tops out around 245 concurrent relay clients — far beyond target deployment size, and the DB is fresh per deployment.
- **`frps_auth_token` is a scanner filter, not an independent trust layer.** It rides the same registration response any `register_token` holder can trigger, so it inherits that token's scope. It still earns its place: it stops internet scanners on the exposed control port and blocks proxy-name squatting. The trade-off is that `register_token` becomes the deployment's crown jewel.
- **Deploy-time selection, not runtime fallback** between `mesh_overlay` and `relay`. Auto-detecting "was this failure real or transient?" is the same category of judgment that already produced one open bug in this codebase; an operator can redeploy with a different `connectivity` value.
- **`frp` is kept an internal implementation detail** — no class name, config key, or `provider_metadata` key names it (only the `requires_frp_check` flag and `relay.py`'s own comments), consistent with how `mesh_overlay` avoids naming Tailscale.

## Follow-up commit

`c4a72269` fixes a defect found after the initial push: `relay_server_addr` was only checked for non-empty, but `register_client` parses it with `.rpartition(":")` + `int(port)`. Three malformed-but-non-empty values passed validation and then raised `ValueError`, 500ing the first relay registration — including the likeliest mistake, omitting the port entirely (`allocator.example.com`). That was precisely the failure the check's own comment claimed to prevent. It now requires `host:port` with a numeric port in 1–65535, with a test pinning that bracketed IPv6 (`[::1]:7000`) still parses.

## Security fixes (commit `33b7e8a9`)

A self-review of this PR turned up three defects, all now fixed with tests. Worth reading, since one was reachable by any `register_token` holder:

1. **Path traversal → arbitrary file write/delete.** `hostname` was validated only for truthiness; `register_client` returns it verbatim as `client_id`; `relay_manager` interpolated it straight into a path, so `'../../../../tmp/pwned'` escaped the config directory. `start_visitor` created/overwrote that path and `stop_visitor` unlinked it, as the allocator user. Now rejected at the registration boundary (400) **and** re-checked inside `relay_manager`, so the module is safe regardless of caller.
2. **TOML injection via the same field.** `client_id` went into four unquoted TOML string positions; a quote or newline corrupted the config so `frpc` failed to start, leaving that client silently unreachable (per-client visitor liveness isn't in `/api/health`). All interpolated strings now go through `json.dumps`, which also protects against a stray quote in operator-supplied `relay_server_addr`.
3. **Secrets world-readable.** The visitor config (per-client `secretKey`) and `/tmp/frps.toml` (`auth.token`) were both `0644`, while the CLI already writes registration secrets at `0600`. Now `0600` in a `0700` directory, created via `os.open` rather than write-then-chmod; `start.sh` uses `umask 077` in a subshell so it doesn't leak into nginx startup.

Note these are **not** caught by coverage gates — the original tests exercised the vulnerable code with well-formed hostnames only, which is exactly why CI was fully green with the bug present.

Known follow-ups deliberately left out of this PR: the `MAX(octet)+1` alias allocation is a read-then-write race under concurrent registration; there's no upper bound before the octet overflows past 255; `stop_visitor`'s `proc.wait(timeout=5)` can raise `TimeoutExpired` and orphan an untracked process; and `start_visitor` failing after the DB row is committed leaves a registered-but-unreachable client.

## Hand-test results (commits `ebe7d215`, `f21f8f73`)

Exercised for real with two containers on a Docker network: the allocator built from `Dockerfile.dev`, and a fake BYO client running `frpc` in proxy mode with throwaway HTTP services on `:7070`/`:6080`.

**It found a showstopper that no unit test could.** The allocator's per-client visitor is itself an `frpc` client, so it must present the deployment's `auth.token` when logging in to `frps`. It never did, so `frps` rejected every visitor login (`token in login doesn't match token from configuration`) and `frpc` exited immediately — while registration still returned a healthy 200 and `/api/health` still said `frp: ok` (correctly: `frps` was fine, it was the *per-client visitor* that died, and per-client liveness is deliberately not surfaced). **No relay client could ever have been reached.** Unit tests mock `Popen`, so the exit is unobservable, and they only assert the config keys we thought to check.

It also surfaced a second gap: only the prod `Dockerfile` got the `frp` binaries. `Dockerfile.dev` had none — the same parity gap the spec warned about for the client image and then wrongly dismissed for the allocator ("only one variant"). This matters more than it appears, because the prod Dockerfile installs `lablink-allocator-service` **from PyPI**, so `Dockerfile.dev` is the only image that can run unreleased relay code. A corollary worth knowing: CI's image build/verify jobs never exercise this branch's Python.

Verified working after the fixes:

| Check | Result |
|---|---|
| `frps` in-process, gated on `FRPS_AUTH_TOKEN` | running; `/api/health` → `{"frp":"ok"}` |
| Register with `{"relay": true}` | 200 with `relay_secret_key`, `relay_server_addr`, `frps_auth_token`; alias octet 10 |
| Visitor process | authenticates to `frps`, binds `127.0.0.10:6080` and `127.0.0.10:7070` |
| **Byte path** | `curl 127.0.0.10:7070/MARKER` from inside the allocator returns content served by the *client* container — visitor → frps → client frpc → client service |
| DB accessors (real Postgres) | `get_relay_alias` → `10` (int), `allocate_relay_alias_octet` → `11`, `_resolve_relay_alias` → `127.0.0.10`, unknown host → `RotationFailed` |
| Second client | gets octet 11; two independent visitors |
| Unregister | `cleanup_client_identity` kills only that client's visitor and removes only its config; the other survives |
| Traversal hostname | rejected 400; nothing written outside the visitor dir |
| File modes | visitor dir `700`, visitor config `600`, `/tmp/frps.toml` `600` |

Still not covered: the participant-facing leg (browser → nginx `/proxy/<token>` → alias) was not driven, since that needs a full VM-assignment flow with a real KasmVNC on the client. The alias-resolution function feeding it is verified against live Postgres.

## Reviewer notes

Two things I'd especially like eyes on:

1. **`routes/registration.py`'s response block** was converted from a `jsonify(...)` call into a dict so the relay fields can be added conditionally. That call had grown to 15 fields, and an earlier draft of this work would have rebuilt it from a stale 10-field list, silently dropping `repository`, `subject_software`, `startup_max_attempts`, `startup_base_delay_seconds`, and `startup_success_check_b64`. `test_register_response_preserves_all_non_relay_fields` now pins all 15 — please sanity-check the diff there anyway.
2. **`validate_config.VALID_CONNECTIVITY` and `providers/registry._CONNECTIVITY_BUILTIN`** are two independent enumerations of the same set. Adding `relay` to only the registry would leave every unit test green while `connectivity: relay` stayed rejected at config validation — i.e. the feature unreachable in any real deployment. Both are updated here; worth knowing for future connectivity modes.

---

### DIFF
--- a/packages/allocator/Dockerfile
+++ b/packages/allocator/Dockerfile
@@ -62,6 +62,20 @@ RUN for f in /usr/share/kasmvnc/www/assets/ui-*.js; do \
             || { echo "ERROR: sed did not remove the clobber from $f"; exit 1; }; \
     done
 
+# frp (Fast Reverse Proxy) -- relay connectivity's tunnel server (frps)
+# and the per-client frpc-visitor subprocess that pulls tunneled ports
+# down to loopback aliases nginx can dial. See the
+# relay-client-connectivity design spec. Harmless when unused --
+# lan_direct/mesh_overlay deployments never start frps (start.sh gates
+# it on FRPS_AUTH_TOKEN's presence).
+ARG FRP_VERSION=0.60.0
+RUN curl -L -o /tmp/frp.tar.gz \
+      "https://github.com/fatedier/frp/releases/download/v${FRP_VERSION}/frp_${FRP_VERSION}_linux_amd64.tar.gz" && \
+    tar -xzf /tmp/frp.tar.gz -C /tmp && \
+    mv "/tmp/frp_${FRP_VERSION}_linux_amd64/frps" /usr/local/bin/frps && \
+    mv "/tmp/frp_${FRP_VERSION}_linux_amd64/frpc" /usr/local/bin/frpc && \
+    rm -rf /tmp/frp.tar.gz "/tmp/frp_${FRP_VERSION}_linux_amd64"
+
 COPY lablink-nginx.conf /etc/nginx/conf.d/lablink.conf
 
 # Disable nginx's default site so it doesn't compete with lablink-nginx.conf.
--- a/packages/allocator/Dockerfile.dev
+++ b/packages/allocator/Dockerfile.dev
@@ -58,6 +58,21 @@ RUN for f in /usr/share/kasmvnc/www/assets/ui-*.js; do \
             || { echo "ERROR: sed did not remove the clobber from $f"; exit 1; }; \
     done
 
+# frp (Fast Reverse Proxy) -- relay connectivity's tunnel server (frps)
+# and the per-client frpc-visitor subprocess that pulls tunneled ports
+# down to loopback aliases nginx can dial. Must stay in lockstep with the
+# identical block in Dockerfile: this is the image built from local source,
+# so it is the ONLY one that can exercise unreleased relay code (the prod
+# Dockerfile installs lablink-allocator-service from PyPI). Harmless when
+# unused -- start.sh gates frps on FRPS_AUTH_TOKEN's presence.
+ARG FRP_VERSION=0.60.0
+RUN curl -L -o /tmp/frp.tar.gz \
+      "https://github.com/fatedier/frp/releases/download/v${FRP_VERSION}/frp_${FRP_VERSION}_linux_amd64.tar.gz" && \
+    tar -xzf /tmp/frp.tar.gz -C /tmp && \
+    mv "/tmp/frp_${FRP_VERSION}_linux_amd64/frps" /usr/local/bin/frps && \
+    mv "/tmp/frp_${FRP_VERSION}_linux_amd64/frpc" /usr/local/bin/frpc && \
+    rm -rf /tmp/frp.tar.gz "/tmp/frp_${FRP_VERSION}_linux_amd64"
+
 COPY lablink-nginx.conf /etc/nginx/conf.d/lablink.conf
 
 # Disable nginx's default site so it doesn't compete with lablink-nginx.conf.
--- a/packages/allocator/src/lablink_allocator_service/conf/structured_config.py
+++ b/packages/allocator/src/lablink_allocator_service/conf/structured_config.py
@@ -192,6 +192,10 @@ class ManualConfig:
               path through its own nginx, same as the AWS path. For
               clients that aren't on the allocator's LAN (e.g. a
               Run:AI-hosted workload).
+            - "relay": the client dials out to the allocator through an
+              frp tunnel instead of joining any overlay. The fallback of
+              last resort when a client's network won't carry Tailscale
+              at all.
         overlay_tailnet (str): The Tailscale tailnet's MagicDNS domain
             suffix (e.g. "example.ts.net"), used to resolve a registered
             client's overlay hostname, and/or the allocator's own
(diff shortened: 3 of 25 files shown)


7174Andy:

PR #417, opened 2026-07-29
feat(client+cli): relay connectivity client side — frpc proxy + `lablink client register --relay`

7174Andy:
> **Stacked on #416.** Targets `andrew/feat-relay-client-connectivity`, so this diff shows only Plan 2's 8 commits. GitHub will retarget to `main` automatically once #416 merges. **Review #416 first.**

## Summary

Plan 1 (#416) made a relay client *reachable* — allocator-side `frps`, a per-client `frpc` visitor on a loopback alias, and a registration endpoint that mints three tunnel values. Nothing consumed them and there was no way to ask for relay.

This is the client half: `lablink client register --relay` persists those values into the existing mode-`0600` env file, and the client image renders an `frpc` proxy config and dials out.

## Changes

| Area | Change |
|---|---|
| `packages/client/Dockerfile` + `Dockerfile.dev` | `frpc` install — **both**, with a parity test |
| `packages/client/start.sh` | Relay block: renders `/tmp/frpc.toml` at `0600`, launches `frpc`, verifies it survived |
| `packages/cli/.../api.py` | `RegistrationClient.register(relay=True)` sends `provider_metadata={"relay": True}` |
| `packages/cli/.../commands/register.py` | `--relay` validation, routing, env-file values, no host-port publishes |
| `packages/cli/.../app.py` | The `--relay/--no-relay` flag |

Supports **both** run-locally and hand-off modes, mirroring mesh-overlay.

## Verified end to end

Two containers on a Docker network, registration driven through the CLI:

- `curl http://127.0.0.10:6080/` from inside the allocator returned **HTTP 401** — KasmVNC's real auth response, not a connection error — with the allocator's frps log showing `get a user connection` at the same timestamp. Bytes provably traversed the tunnel.
- A deliberately wrong `FRPS_AUTH_TOKEN` made the client **exit 1** with `frpc exited immediately`, confirming a dead tunnel fails loudly instead of reporting healthy.

Suites: CLI **693 passed**, client **152 passed**, allocator **842 passed / 19 skipped**. Lint clean, `bash -n` clean.

## Design decisions worth reviewing

- **Gate on `CONNECTIVITY = relay`, not on a secret's presence.** Presence-gating means a missing secret yields *no tunnel and no error* — the exact shape that made Plan 1 ship broken. A missing secret now fails loudly.
- **`transport.dialServerTimeout = 2`.** Caught by the final review: the 3s liveness check only detected *immediate* frpc death. frpc's dial timeout defaults to 10s, so a **blackholed egress path** — precisely what a Tailscale-hostile network produces — would leave frpc alive at t=3s, pass the check, and report healthy before dying. The 2s timeout brings that failure inside the window, and a test asserts the two numbers *relatively* so neither can drift past the other.
- **Process substitution, not a pipeline, for the frpc launch.** After a pipeline `$!` is `sed`'s PID, which survives regardless — the liveness check would have verified nothing. A test pins this.
- **Proxy names are exactly `$VM_NAME-kasmvnc` / `$VM_NAME-agent`**, matching the allocator visitor's `serverName`. A mismatch produces no error on either side, so it gets its own test.
- **`KASMVNC_LISTEN=127.0.0.1` and no `--publish`** for relay clients: frpc dials loopback inside the container, so there is no reason to expose ports on a network we chose not to trust. Publishing would also leave a *dead* `6080`, since Docker forwards to the container's external interface.
- **No new container capabilities.** Unlike Tailscale (`NET_ADMIN`, `/dev/net/tun`, state volume), frp needs none; a test asserts their absence so nobody copies the overlay branch wholesale.

## Known limitations

- **`frpc` is not supervised after startup.** The launch is verified; a later death is not detected. The allocator-side per-client liveness check remains the real fix and is still deferred.
- **Plan 3 is still outstanding**, so a relay deployment needs a hand-edited `config.yaml` *and* a hand-set `FRPS_AUTH_TOKEN` on the allocator container: `deploy_compose.py` has no relay wiring and the wizard cannot select it.
- The participant-facing leg (browser → nginx `/proxy/<token>` → alias) is still unexercised.
- Two minors deliberately deferred: a `KeyError` on a partial relay payload (unreachable against any Plan-1+ allocator, which sets all three fields together) and unescaped TOML interpolation in `start.sh` (all values allocator-minted or hostname-regex-validated; a stray quote fails loudly rather than injecting).

## Note for reviewers

7 CLI tests fail under a colour-capable terminal because they assert plain substrings against Rich output containing ANSI codes. **Pre-existing** — they fail identically at the base commit — and invisible in CI, which runs without a TTY. Unrelated to this branch, but worth knowing before a local run alarms you.

---


7174Andy:

PR #418, opened 2026-07-30
feat(cli): relay connectivity deployment side — `lablink deploy`/`configure` wiring

7174Andy:
Plan 3 of 3 for relay connectivity, stacked on #417 (which is stacked on #416). **Review/merge #416 → #417 → this.** Diff shown against `andrew/feat-relay-client-plan2`; only the two commits here are new.

## Summary

- `lablink deploy` now renders the `frps` env + published control port for a `connectivity: relay` deployment, so it no longer needs a hand-set `FRPS_AUTH_TOKEN` on the container.
- `lablink configure` can now produce a relay config at all — it gains the third connectivity radio, a `relay_server_addr` input, and mints the `frps_auth_token` the operator has no way to supply.
- Together these close the last "hand-edited `config.yaml`" gap called out in #416/#417: relay is now reachable through the normal configure → deploy → `register --relay` path.

## Changes Made

**`deploy_compose.py`**
- `.env` gains `FRPS_AUTH_TOKEN` + `FRPS_BIND_PORT` when connectivity is relay. That file is already mode `0600`, which is where a deployment secret belongs — the override references it rather than inlining the token.
- New `templates/docker-compose-relay-override.yml`, rendered as `docker-compose.override.yml`, supplying the container env and publishing `${FRPS_BIND_PORT}:${FRPS_BIND_PORT}`. Deleted whenever connectivity isn't relay.
- Two preflights: empty `relay_server_addr`, and weak/empty `frps_auth_token` (via Plan 1's `is_weak_frps_token`).
- `_print_summary` treats relay as off-LAN alongside `mesh_overlay` (prefers the Funnel URL, shows the `--no-run-locally` note), prints the valueless `--relay` register command, and names both the published port and the address clients dial.

**`tui/wizard.py`**
- Third `relay` radio + `relay_server_addr` input on `ManualConnectivityScreen`, shown unconditionally like the tailnet field above it — the validator enforces "required for relay", so no show/hide wiring.
- `secrets.token_urlsafe(32)` minted into `frps_auth_token` on Next, only when the value on record is weak.
- Error filter extended to `relay_server_addr` / `frps_auth_token` so those validator messages actually reach the screen's error label.

## Testing

712 CLI tests pass (+27 new), `ruff check src tests` clean.

- `TestRenderComposeDirRelay` — env lines carry token/port, token never lands outside `.env`, override sets container env + publishes the port, non-relay writes neither, switching away deletes a stale override, relay keeps the sidecar-free base template.
- `TestRelayPreflight` — missing address and each flavour of weak token rejected; relay is *not* caught by the Tailscale-authkey gate (it exists for networks that can't run Tailscale).
- `TestPrintSummaryRelay` — valueless `--relay` with no overlay flags, published port + dialled address reported, Funnel URL preferred over the LAN IP.
- Wizard: relay writes the address and a strong token, missing *and* malformed addresses block progression, an existing strong token survives, a weak one is replaced, non-relay mints nothing.

**Verified against the real `docker compose` binary**, not just asserted in tests: a rendered relay workdir's `docker compose config` merges to both `80:5000` and `7100:7100` with `FRPS_AUTH_TOKEN`/`FRPS_BIND_PORT` substituted from `.env` — for the plain base template and for the Funnel/sidecar one.

Not covered: a live `configure → deploy → register --relay` run. That needs an allocator image containing #416's frp install, and the published `linux-amd64-latest` predates it (`pull_policy: always` blocks substituting a local dev image without hand-editing the rendered compose file).

## Design Decisions

- **An override file, not a third base template.** The base template already forks on whether a Tailscale sidecar is needed, and relay is orthogonal to that fork — a variant per combination is four files. Compose auto-loads `docker-compose.override.yml` from the project directory (which is what `docker compose up` runs in) and *concatenates* `ports`, so relay adds its port without touching either base template. The cost is that the override must be deleted when connectivity changes, which `render_compose_dir` does and a regression test pins.
- **The token is generated, not collected.** It is machine-to-machine only, and Plan 1's validator rejects weak values — collecting it would deadlock the ReviewScreen on a field the operator can't usefully fill. Regenerating on every `configure` run would silently break already-registered clients, hence generate-only-if-weak.
- **Deploy-time preflights duplicate `get_config_errors` on purpose.** `lablink deploy` never calls that validator for the manual provider, and the allocator doesn't validate its config at startup — so for a hand-edited `config.yaml` this is the only gate. Same reasoning as the existing `lan_direct` + `tailscale_funnel` preflight next to it.
- **Publishing the port is where this stops.** Making `manual.relay_server_addr` route to the published port (port-forward, VPN, public IP) stays the operator's choice, as #416's design states — and Funnel is not an option for it, since Funnel carries HTTPS, not raw TCP. The summary says so rather than the code pretending to automate it.
- **No docs changes.** Nothing under `docs/` documents `mesh_overlay` either; connectivity modes are spec-documented only.

## Related Issues

Completes the relay connectivity roadmap: #416 (allocator foundation) → #417 (client + `register --relay`) → this (deployment wiring).

---

### DIFF
--- a/packages/cli/src/lablink_cli/commands/deploy_compose.py
+++ b/packages/cli/src/lablink_cli/commands/deploy_compose.py
@@ -55,6 +55,9 @@ FUNNEL_ENABLE_RETRY_DELAY_SECONDS = 2
 # dependencies, so a cross-package import would fail there. Guarded by
 # test_deploy_compose.py::TestCanonicalUrlFile::test_filename_matches_allocator.
 CANONICAL_URL_FILENAME = "allocator-url"
+# Compose's own conventional override filename — auto-loaded from the project
+# directory with no `-f` needed. Only written when connectivity is "relay".
+RELAY_OVERRIDE_FILENAME = "docker-compose.override.yml"
 
 console = Console()
 
@@ -157,6 +160,19 @@ def render_compose_dir(
     template = resources.files("lablink_cli.templates").joinpath(template_name)
     (target / "docker-compose.yml").write_text(template.read_text())
 
+    # 1b. Relay adds an override rather than a third base template (see the
+    #     override file's own header). Deleted when connectivity isn't relay,
+    #     so a redeploy that switches away stops publishing frps's port —
+    #     `docker compose up` would otherwise keep honouring a stale override.
+    override_path = target / RELAY_OVERRIDE_FILENAME
+    if cfg.manual.connectivity == "relay":
+        override_template = resources.files("lablink_cli.templates").joinpath(
+            "docker-compose-relay-override.yml"
+        )
+        override_path.write_text(override_template.read_text())
+    else:
+        override_path.unlink(missing_ok=True)
+
     # 2. Render .env — only the values the compose template substitutes.
     #    No DB or admin creds here: they're inside config.yaml. Read the
     #    OLD .env (if any) before overwriting it, so a redeploy that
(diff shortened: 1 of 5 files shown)


7174Andy:

PR #435, opened 2026-08-06
fix(cli): make metrics export work on the manual/BYO path

7174Andy:
## Summary

- `lablink stats` and `lablink export-metrics --client` could not run at all on a manual/BYO deployment — allocator URL and admin credentials were resolved AWS-only.
- `export-metrics --allocator` exported the *entire* local cache, so a file named for one deployment carried every other deployment's rows.
- The compose deploy path never recorded deployment metrics in the first place, so there was nothing for a BYO operator to export.

## Changes Made

**1. Manual-provider resolution** (`utils.py`, `status.py`, `stats.py`)

`get_allocator_url()` resolves only from Terraform outputs (`ec2_public_ip`) or `cfg.dns.domain`; a compose stack has neither, so it returned `""` and both metrics commands exited with *"Could not determine allocator URL."* `status` and `logs` had each grown their own manual branch — the two metrics commands never did.

- Manual short-circuits to `http://localhost:<HTTP_PORT>`, the port both compose templates publish and the same base URL `deploy_compose._health_poll` already polls.
- `resolve_admin_credentials()` reads the rendered compose workdir's `config.yaml` instead of the AWS deploy dir (which never exists there, so every BYO operator got the interactive prompt).
- Extracted `resolve_from_saved_config(path)` so both providers share one yaml reader; `status._resolve_manual_admin_credentials` delegates to it instead of duplicating the read (-11 lines).
- `print_admin_credentials_hint(cfg)` names the compose `config.yaml` on a manual 401, not a deploy dir absent from that machine.

**2. Scoped allocator export** (`export_metrics.py`, `deployment_metrics.py`, `deploy.py`, `app.py`)

The cache is global — one file per deploy attempt, never pruned — and the export wrote all of it. Filtering by `deployment_name` alone was insufficient: a name can be reused across providers, and a `sleap-lablink` config switched from `aws` to `manual` still matched 37 old AWS Terraform records.

- Records now carry `provider`; readers default a missing value to `"aws"`, which every pre-existing record is.
- Export filters on `deployment_name` **and** `provider`. An unscoped export announces how many deployments it spans.
- `--allocator` alone now loads the config when one exists (it deliberately skipped it before), so scoping applies to the invocation a BYO operator actually reaches for. No config at all still exports unscoped — that is what kept the command working after `lablink destroy`.

**3. Compose deploy records metrics** (`deploy_compose.py`, `deployment_metrics.py`)

`run_deploy_compose` now times `docker compose up -d` and the health poll, then writes status/total. New `allocator_compose_up_duration_seconds` field — kept separate from the terraform-apply timing so a compose deploy does not look like a Terraform one in the CSV. `region`/`template_version` stay `None` rather than being invented.

## Testing

763 CLI tests pass; `ruff check` clean across all three packages. No source outside `packages/cli` is touched.

New coverage:
- `get_allocator_url` manual short-circuit, asserting it never consults `get_deploy_dir`
- manual creds read from the compose workdir, asserting the AWS deploy dir is not consulted
- hint wording for aws / manual / no-cfg
- export scoped by deployment; the reused-name-across-providers case with its AWS counterpart (legacy records with no `provider` field still export under an AWS config)
- compose deploy: success record, health-timeout → `failed`, declined confirmation → no record

Manual testing against a live compose stack:
- `export-metrics --client` reached the allocator (`200 OK`, `{"count": 0, "vms": []}`) where it previously died on URL resolution; credentials resolved with stdin closed, so no prompt
- `export-metrics --allocator` on the real cache: was 96 rows spanning 3 deployments → now 1 row, `provider=manual`, compose up 3.26s, health check 9.05s, total 12.31s, `status=success`, empty `terraform_*` columns

## Design Decisions

**`localhost` over the canonical public URL for manual.** Both compose templates publish `${HTTP_PORT}:5000` on the host, and the CLI's manual paths already assume they run on that host (`status`/`logs` shell into the local container). Correct even behind Funnel/Cloudflare, since the tunnel dials that same published port.

**`provider` on the record, not inferred at read time.** Nothing in a stored record distinguishes an AWS deploy from a compose one except which timing columns are populated — too fragile to key filtering on.

**Compose path catches `SystemExit`; the AWS path deliberately does not.** In `deploy.py`, `SystemExit` means the operator cancelled, and leaving the record `in_progress` is honest. In `run_deploy_compose`, `_compose_up` and `_health_poll` both raise it for *genuine* failures (non-zero compose exit, health timeout), so mirroring AWS would under-report real failures. `KeyboardInterrupt` still escapes both.

**Compose record is created after the confirmation gate.** A declined "Proceed?" leaves no `in_progress` file at all. The AWS path creates its record earlier and does litter — 27 of 97 records in one real cache were `in_progress` leftovers.

## Incidental fixes

Two test-isolation bugs surfaced while doing this, both in `conftest.py`:

- `mock_cfg.provider` was a `MagicMock`: truthy, never equal to any provider string, and not JSON-serializable. It was masking exactly the class of bug fixed here. Set to `"aws"`, matching the real `Config` default.
- Added an autouse `no_real_deployment_cache` fixture (sibling to the existing `no_real_docker` guard). `DEPLOYMENTS_DIR` is a module-level constant, so once the compose path recorded metrics the suite wrote 25 `testlab-*.json` files into the developer's own cache, where they then appear in `export-metrics` output as real deployments. Only three AWS tests had ever patched that constant by hand; every other deploy test silently did not.

## Known gaps (not in this PR)

- The compose destroy path has no export-before-destroy prompt, unlike `deploy.py`, and `docker volume rm` takes the Postgres volume with all session metrics.
- `bulk_seal_session_metrics()` is called only from the AWS destroy paths; BYO withdrawal (`DELETE /api/v1/clients/<id>`) hard-deletes the row, so metrics vanish unsealed.

---

https://claude.ai/code/session_017WswonxBxQWGrr4itfRYLR

### DIFF
--- a/docs/cli/managing-deployments.md
+++ b/docs/cli/managing-deployments.md
@@ -61,8 +61,20 @@ Writes deployment metrics to disk for offline analysis. Two sources:
 | Flag | Data | Requires allocator running? |
 |---|---|---|
 | `--client` | Per-VM metrics pulled from the allocator's API (boot time, health status, logs) | Yes |
-| `--allocator` | Per-deploy metrics from the local cache at `~/.lablink/deployments/` (deploy duration, Terraform phase timings) | **No** — works after `lablink destroy` |
+| `--allocator` | Per-deploy metrics from the local cache at `~/.lablink/deployments/` (deploy duration, plus Terraform phase timings on AWS or `docker compose up` timing on the manual provider) | **No** — works after `lablink destroy` |
 | *(no flag)* | Both | Yes |
+
+Allocator metrics are **scoped to the deployment and provider in your config**.
+The cache is shared by every deployment you have ever run, so an unscoped export
+would put other deployments' rows in a file named after this one. Point
+`--config` at a different config to export that deployment instead. On a machine
+with no config at all, `--allocator` exports the whole cache and says how many
+deployments it spans.
+
+Reusing one `deployment_name` across providers is why the provider is part of the
+scope: a name deployed first on AWS and later with `provider: manual` has records
+of both shapes in the cache, and the Terraform phase columns say nothing about a
+compose stack.
 
 Other flags:
 
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -373,14 +373,21 @@ Writes metrics to disk for offline analysis. Two data sources:
 With no source flag, both are exported and `_client` / `_allocator` suffixes are
 appended to the base output name.
 
+Allocator metrics are scoped to the config's `deployment_name` **and** provider —
+the cache holds a record per deploy attempt for every deployment on the machine,
+and a name reused across providers has records of two different shapes in it. AWS
+deploys populate the `allocator_terraform_*` columns; manual (compose) deploys
+populate `allocator_compose_up_duration_seconds` instead. Both populate
+`allocator_health_check_duration_seconds` and the total.
+
 | Option | Description |
 |---|---|
 | `--client` | Export per-VM client metrics from the allocator. |
-| `--allocator` | Export per-deploy allocator metrics from the local cache. Skips the network. |
+| `--allocator` | Export per-deploy allocator metrics from the local cache, scoped to this config's deployment and provider. Skips the network. |
 | `-f`, `--format FMT` | `csv` (default) or `json`. |
 | `-o`, `--output PATH` | Output file path. With a single source flag it is the literal path; with both (or neither), it is a base name and `_client` / `_allocator` suffixes are added before the extension. Default: `metrics_client.<fmt>` and/or `metrics_allocator.<fmt>`. |
 | `--include-logs` | Include `cloud_init_logs` and `docker_logs` columns. Large — opt-in only. |
-| `-c`, `--config PATH` | Path to `config.yaml`. Skipped when only `--allocator` is passed. |
+| `-c`, `--config PATH` | Path to `config.yaml`. With only `--allocator`, it is loaded when present (to scope the export) but not required — the command still works on a machine that has no config, exporting the whole cache. |
 
 ---
 
--- a/packages/cli/src/lablink_cli/app.py
+++ b/packages/cli/src/lablink_cli/app.py
@@ -767,8 +767,10 @@ def export_metrics(
         False,
         "--allocator",
         help=(
-            "Export per-deploy allocator metrics from the local cache. "
-            "Works without a running allocator (e.g. after `lablink destroy`)."
+            "Export per-deploy allocator metrics from the local cache, "
+            "scoped to this config's deployment_name. Works without a "
+            "running allocator (e.g. after `lablink destroy`); passed "
+            "alone it loads no config and exports every deployment."
         ),
     ),
     config: str = typer.Option(
@@ -781,15 +783,24 @@ def export_metrics(
     """Export deployment metrics to CSV or JSON.
 
     Pass --client for per-VM metrics from the allocator. Pass --allocator
-    for per-deploy metrics from the local cache. With no flag, exports
-    both. Passing only --allocator skips the network entirely.
+    for per-deploy metrics from the local cache, scoped to this config's
+    deployment_name. With no flag, exports both. Passing only --allocator
+    skips the network entirely.
     """
     from lablink_cli.commands.export_metrics import run_export_metrics
 
-    # Skip config load when only --allocator is requested — it doesn't need
-    # the config and we want this command to work even after `lablink destroy`.
-    needs_cfg = client or not allocator
-    cfg = _load_cfg(config) if needs_cfg else None
+    # --client cannot work without the config (allocator URL + admin creds),
+    # so a missing one is fatal there. --allocator does not need it, but load
+    # it when it's there anyway: the config's deployment_name is what scopes
+    # the cache export to this deployment instead of dumping every deployment
+    # the operator has ever run. Only a machine with no config at all falls
+    # through to None (unscoped), which keeps the command usable after a wipe.
+    config_path = Path(config) if config else DEFAULT_CONFIG
+    cfg = (
+        _load_cfg(config)
+        if client or not allocator or config_path.exists()
+        else None
+    )
 
     run_export_metrics(
         cfg,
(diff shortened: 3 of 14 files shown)


7174Andy (commented):
## Code Review

### Summary

Three chained fixes to the metrics path for manual/BYO deployments: resolution (URL + creds), export scoping (deployment + provider), and recording (compose deploy). The chain is coherent and each commit stands on its own. Verified end-to-end against a live compose stack, which is what gives the scoping numbers credibility.

No correctness bugs in the resolution or scoping logic. Two things worth addressing before merge: an `in_progress` window in the new recording path that undercuts the PR's own stated design goal, and missing docs updates that CLAUDE.md requires.

### Strengths

- `resolve_from_saved_config` extraction genuinely removes duplication rather than adding a layer — `status._resolve_manual_admin_credentials` drops 11 lines and both providers now share one yaml reader.
- The `provider` field is stored rather than inferred. Inferring from "which timing columns are populated" would have been the tempting shortcut and would rot the first time a column changes.
- `no_real_deployment_cache` fixes an actual live bug, not a hypothetical one: the suite was writing into the developer's own `~/.lablink/deployments/`, where the records then surface in real `export-metrics` output.
- Claims are backed by measurements (88-row export with 59 foreign rows; 96 → 1 row after scoping) rather than assertions.

### Issues Found

#### Important

**1. `in_progress` window in `run_deploy_compose` — `deploy_compose.py:490-528`**

The record is written at L490, but the `try/except` only wraps L513-518 (`_compose_up`, `_health_poll`). Two stretches sit outside it:

- L492 `render_compose_dir(...)` — writes several files and calls `save_config`; an `OSError` or a missing packaged template raises here.
- L542-549 `_disable_funnel()` / `_enable_funnel()` / `_write_canonical_url(...)` — the last one writes a file, so `OSError` is live.

A failure in either leaves `status=in_progress` permanently, with null timings, indistinguishable from a Ctrl-C. That is precisely the litter the commit message claims this path avoids ("27 of 97 records in one real cache were in_progress leftovers"). The declined-confirmation claim is accurate; the general one is weaker than stated.

Fix is mechanical — widen the `try` to cover L492 through L549, leaving the success write where it is.

**2. Docs not updated (CLAUDE.md: "Update the relevant page under `docs/` if behavior changes")**

Three user-visible behavior changes, none documented:

- `docs/cli/managing-deployments.md:64` still describes `--allocator` as *"Per-deploy metrics from the local cache … (deploy duration, Terraform phase timings)"*. Manual deploys now record too, and their timings are not Terraform ones.
- Neither that table nor `docs/reference/cli.md:371` mentions that the export is now scoped to the config's deployment and provider — the single most surprising change for an existing user.
- The new `allocator_compose_up_duration_seconds` column is undocumented.

Worth noting the flip side: `docs/cli/byo-clients.md:198-199` already claims *"Every other command, `stats` and `export-metrics` included, behaves as it does on the AWS path."* That was false before this PR and is true after it — the bug contradicted documented behavior, which strengthens the case for these being fixes rather than enhancements.

#### Minor / Suggestions

**3. A capability was removed without an escape hatch — `app.py:791-802`**

With a config present, there is no longer any way to get the unscoped whole-cache export; `--allocator` used to do exactly that. The only remaining route is running on a machine with no config at all. Deliberate and described in the PR, but if anyone was using it for cross-deployment analysis it breaks silently. `--all-deployments` would restore it in ~3 lines; I'd hold off until someone asks.

**4. Asymmetric None-guarding — `export_metrics.py:200-205`**

`deployment_name=getattr(cfg, "deployment_name", None)` vs `provider=getattr(cfg, "provider", None) if cfg else None`. Both are correct (`getattr(None, "x", None)` returns `None`), but the asymmetry suggests one path is unsafe when neither is. Pick one form.

**5. Naming consistency — `utils.py`**

New `resolve_from_saved_config` is public while its only sibling `_resolve_from_deploy_dir` stays private. Cross-module use from `status.py` justifies dropping the underscore; a one-line note saying so would stop the next reader from "fixing" it.

**6. Two helpers now disagree on scheme — `utils.py:307` vs `status.py:~800`**

`get_allocator_url`'s manual branch hardcodes `http`, while `_run_status_manual` picks `https` when `ssl.provider == "self_signed"`. Unreachable today (`SUPPORTED_SSL_FOR_MANUAL = ("none",)` makes deploy refuse anything else), so this is latent, not a bug — but it is exactly the kind of divergence that bites when someone relaxes that constraint.

### Recommendations

1. Widen the `try` in `run_deploy_compose` to cover render and the exposure block (Important #1).
2. Update `docs/cli/managing-deployments.md` and `docs/reference/cli.md` for scoping, manual recording, and the new column (Important #2).
3. Normalize the `getattr` guards and add the one-line note on the public helper.
4. Confirm CI before merge — it has not reported yet, and it runs Python 3.10 in per-package isolated envs while local verification was 3.12 on the shared workspace venv.

### Questions

- Should the AWS path adopt the compose path's "create the record after the confirmation gate" ordering? It is the source of all 27 `in_progress` leftovers, and the asymmetry between the two paths is now a documented deliberate difference rather than an accident.
- Is default-to-`"aws"` for pre-existing records the intended long-term behavior, or should they be backfilled once? Default is correct today; it becomes a silent wrong answer only if a *third* provider ever writes records without setting the field.

7174Andy:
Both Important findings addressed in 6f08cc9.

**1. `in_progress` window closed.** The `try` now spans from `render_compose_dir` through the exposure block, so the two stretches that previously escaped past the failure write — render (writes files, calls `save_config`) and `_write_canonical_url` — record `status=failed` instead of stranding the record. The success write stays outside the `try`. New test drives an `OSError` out of `render_compose_dir` and asserts `failed` + the error text + null phase timings.

**2. Docs updated.**
- `docs/cli/managing-deployments.md`: `--allocator` no longer described as Terraform-timings-only; added the scoping rule and why the provider is part of the scope.
- `docs/reference/cli.md`: same scoping note, which timing columns each provider populates (including the new `allocator_compose_up_duration_seconds`), and fixed the now-false *"config is skipped when only `--allocator` is passed"* — it is loaded when present, just not required.

Verified: 764 CLI tests pass, `ruff check` clean, `mkdocs build` emits no warnings from either edited page.

Minor findings #3-#6 left as-is deliberately: #3 (`--all-deployments`) until someone asks, #4-#6 are cosmetic/latent. Happy to fold them in if you'd rather they not linger.

⚠️ CI still has not triggered on this branch — 0 workflow runs and 0 check-runs for `6f08cc9`, despite `ci.yml` matching `packages/**`. Worth a look at the Checks tab; my token can't read the repo's Actions policy to say why.

